PR #17, opened 2022-07-25
Doc: initial draft of test specification

niklaslong:
The spec doesn't contain any specific test cases yet, these will be documented as we map out the protocol.


ljedrz (approved):
LGTM

PR #18, opened 2022-08-02
fix: wait for node stopping / starting

ghost:


ljedrz (commented):
LGTM squashed :+1:

src/tests/conformance/handshake.rs
```diff
@@ -18,9 +18,10 @@ async fn handshake_when_node_receives_connection() {
         listener_ip: Some("127.0.0.1".parse().unwrap()),
         ..Default::default()
     };
-
-    // TODO: replace with a connection from the node to signal readiness.
-    tokio::time::sleep(CONNECTION_TIMEOUT).await;
+    wait_until!(
```
niklaslong, line 21: We might want to drop the connection on success, just so we don't have a persistant connection during tests?

lwojtow: Isn't it dropped when inside `loop` ends?

niklaslong: Yep, my bad, the stream will get dropped 👍 . That being said, I think this call should probably not be in the test but in the node setup code (as with the Zcash impl).

niklaslong: Note that in the zcash impl, we do the opposite, we wait for a connection from the node—I currently don't see a problem with your approach, just that it may cause breakage on the node's side if it can't handle too many conns being opened (1 per ms if I recall currectly in the `wait_until!` impl).

niklaslong: > Yep, my bad, the stream will get dropped
Well, it _should_ but `TcpStream` doesn't actually implement `Drop`. The read half is usually dropped when the write half is but since we don't actually split the stream, it might be best to call `shutdown` to be safe.

src/setup/node.rs
```diff
@@ -85,7 +85,7 @@ impl Node {
                 }
                 Some(exit_code) => Some(format!("crashed with exit code {}", exit_code)),
             };
-
+            child.wait()?;
```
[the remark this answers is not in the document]
lwojtow: Here's what happened: 
Test always passed when run only this one and occasionally failed with 'connection refused' at `synth_node.node().connect(node.addr()).await.unwrap();` when I run all tests quickly multiple times. (So in other words: connect was fine first, then refused later). I figured, that's because some delay before OS kills child process from previous test run, but I'm willing to accept other explanation.

src/setup/node.rs
```diff
@@ -153,7 +172,10 @@ mod tests {
     async fn start_stop_node() {
         let mut node = Node::new().unwrap();
 
-        node.log_to_stdout(true).start().unwrap();
+        node.log_to_stdout(true)
+            .start(Duration::from_secs(2))
```
niklaslong, line 176: Suggestion: wondering if we don't just want this to be a global constant, this isn't the only test we'll be waiting on a node to start and I don't really see why the needed duration would change between different tests?

lwojtow: Similarly, this can be just inside `start()`. What do you think?

src/tests/conformance/handshake.rs
lwojtow: Done

PR #21, opened 2022-08-08
refactor for inner node, add ping-pong basic test

lwojtow:
1. Changed SyntheticNode to InnerNode, add SyntheticNode on top (to hold receiver)
2. Added channel to communicate InnerNode -> SyntheticNode (this matches structure in zcash tests)
3. Added basic test for ping -> pong.

### DIFF
--- a/src/tests/conformance/query/basic.rs
+++ b/src/tests/conformance/query/basic.rs
@@ -1,0 +1,55 @@
+//! Contains test with basic queries.
+//!
+//! Queries and expected replies:
+//!
+//!     - Ping           -> Pong
+
+use crate::{
+    protocol::{
+        codecs::binary::{BinaryMessage, Payload},
+        proto::TmPing,
+    },
+    setup::node::Node,
+    tools::synth_node::SyntheticNode,
+};
+
+#[tokio::test]
+async fn ping() {
+    let mut node = start_node().await;
+    let mut synth_node = start_synth_node().await;
+    synth_node.connect(node.addr()).await.unwrap();
+    let payload = Payload::TmPing(TmPing {
+        r#type: 0,
+        seq: Some(42),
+        ping_time: None,
+        net_time: None,
+    });
+    synth_node.unicast(node.addr(), payload).unwrap();
+    let check = |m: &BinaryMessage| {
+        matches!(
+            &m.payload,
+            Payload::TmPing(TmPing {
+                r#type: 1,
+                seq: Some(42u32),
+                ..
+            })
+        )
+    };
+    assert!(synth_node.expect_message(check).await);
+    synth_node.shut_down().await;
+    node.stop().unwrap();
+}
+
+async fn start_synth_node() -> SyntheticNode {
+    let node_config = pea2pea::Config {
+        listener_ip: Some("127.0.0.1".parse().unwrap()),
+        ..Default::default()
+    };
+    SyntheticNode::new(node_config).await
+}
+
+async fn start_node() -> Node {
+    let mut node = Node::new().unwrap();
+    node.log_to_stdout(false).start().await.unwrap();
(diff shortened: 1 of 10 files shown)


src/tests/conformance/query/basic.rs
```diff
@@ -0,0 +1,55 @@
...
+        r#type: 0,
+        seq: Some(42),
+        ping_time: None,
+        net_time: None,
+    });
+    synth_node.unicast(node.addr(), payload).unwrap();
+    let check = |m: &BinaryMessage| {
+        matches!(
+            &m.payload,
+            Payload::TmPing(TmPing {
```
ljedrz, line 31: shouldn't this be `TmPong` instead?

ljedrz: ah no, I see it's the same message type :+1:

src/tests/conformance/query/basic.rs
```diff
@@ -0,0 +1,55 @@
...
+                r#type: 1,
```
ljedrz, line 32: (this indicates that it's a reply)

src/tests/conformance/query/basic.rs
```diff
@@ -0,0 +1,55 @@
...
+};
+
+#[tokio::test]
+async fn ping() {
+    let mut node = start_node().await;
+    let mut synth_node = start_synth_node().await;
+    synth_node.connect(node.addr()).await.unwrap();
+    let payload = Payload::TmPing(TmPing {
+        r#type: 0,
+        seq: Some(42),
```
ljedrz, line 23: ideally, this would be a random value

src/tools/synth_node.rs
```diff
@@ -0,0 +1,87 @@
...
+    pub fn num_connected(&self) -> usize {
+        self.inner.node().num_connected()
+    }
+
+    pub fn is_connected_ip(&self, addr: IpAddr) -> bool {
+        self.inner.is_connected_ip(addr)
+    }
+
+    pub async fn expect_message(&mut self, check: fn(&BinaryMessage) -> bool) -> bool {
+        let d = Duration::from_secs(20);
```
ljedrz, line 75: this should probably be a highly visible `const`

ljedrz:
It would be best if the new test was in its own commit.

src/tests/conformance/query/basic.rs
lwojtow: No such thing. Pong comes back with different r#type.

src/tests/conformance/query/basic.rs
niklaslong: Might be good to make a comment of this in the test, it's documented in the `.proto` file but anyone reading the test might enjoy this being made explicit here?

src/tools/synth_node.rs
niklaslong: I'd even suggest starting a module with the various timeouts/constants used—makes it easier to tweak (often these are hardware-dependent).

src/tests/conformance/query/basic.rs
```diff
@@ -0,0 +1,55 @@
...
+    protocol::{
+        codecs::binary::{BinaryMessage, Payload},
+        proto::TmPing,
+    },
+    setup::node::Node,
+    tools::synth_node::SyntheticNode,
+};
+
+#[tokio::test]
+async fn ping() {
```
niklaslong, line 17: For each test (or collections of cases depending on what makes sense), it would be good to add an entry to `SPEC.md` and referencing it here.
```suggestion
async fn ping() {
// ZG-CONFORMANCE-<insert case number here>
```

niklaslong (commented):
Looks good to me overall, I've added a few comments/suggestions.

PR #24, opened 2022-08-08
Move node convenience  function from 'common' to respective nodes.

lwojtow:


src/setup/node.rs
```diff
@@ -36,6 +36,15 @@ impl Node {
         })
     }
 
+    pub async fn create_and_start(peers: Vec<SocketAddr>) -> Result<Self> {
```
ljedrz, line 39:
```suggestion
    pub async fn start_with_peers(peers: Vec<SocketAddr>) -> Result<Self> {
```

src/tools/synth_node.rs
```diff
@@ -35,6 +35,15 @@ impl SyntheticNode {
         Self { inner, receiver }
     }
 
+    /// Starts node bound to localhost and with default [pea2pea::Config]
+    pub async fn start_on_localhost() -> Result<Self, AddrParseError> {
```
ljedrz, line 39:
```suggestion
    pub async fn start() -> Result<Self, AddrParseError> {
```
it will always be started on localhost in these tests, so no need to be explicit here

src/setup/node.rs

src/tools/synth_node.rs
lwojtow: Done

ljedrz (approved):
LGTM :+1:

PR #28, opened 2022-08-16
Update derives to satisfy clippy

lwojtow:


ljedrz (approved):
LGTM; this is an auto-generated file, but I'll be happy to have the CI pacified without going through re-generation which might not yet be aware of this lint

PR #30, opened 2022-08-17
Refactor: extract common query/response test code

lwojtow:


src/tests/conformance/query/ping_pong.rs
```diff
@@ -1,5 +1,5 @@
 //! Contains test with ping queries.
-//!
+//! ZG-CONFORMANCE-003
```
ljedrz, line 2: I'm all for moving the spec test names somewhere above, but we should probably be consistent about it; @niklaslong?

src/tests/conformance/mod.rs
```diff
@@ -1,2 +1,27 @@
+use crate::{
+    protocol::codecs::binary::{BinaryMessage, Payload},
+    setup::node::Node,
+    tools::synth_node::SyntheticNode,
+};
+
 mod handshake;
 mod query;
+
+async fn perform_query_response_test(payload: Payload, check: &dyn Fn(&BinaryMessage) -> bool) {
```
ljedrz, line 10:
```suggestion
async fn perform_query_response_test(query_msg: Payload, response_check: &dyn Fn(&BinaryMessage) -> bool) {
```

src/tests/conformance/mod.rs
```diff
@@ -1,2 +1,27 @@
...
+    // Start Ripple node
+    let mut node = Node::start_with_peers(vec![]).await.unwrap();
+
+    // Start synth node and connect to Ripple
+    let mut synth_node = SyntheticNode::start().await.unwrap();
+    synth_node.connect(node.addr()).await.unwrap();
+
+    // Send message if payload given
```
ljedrz, line 18:
```suggestion
    // Send the query message
```

src/tests/conformance/mod.rs
```diff
@@ -1,2 +1,27 @@
...
+    synth_node.unicast(node.addr(), payload).unwrap();
+
+    // Wait for required response
```
ljedrz, line 21:
```suggestion
    // Wait for a response and perform the given check for it
```

src/tests/conformance/mod.rs
lwojtow: Done

src/tests/conformance/mod.rs
lwojtow: Done

src/tests/conformance/mod.rs
lwojtow: Done

src/tests/conformance/query/ping_pong.rs
niklaslong: So long as the codebase is easily searchable and it is clear what cases the comment identifies in the file, I'm fine with either or even a combination.

ljedrz (approved):
LGTM :+1:

PR #31, opened 2022-08-17
Feature: Add test for TmGetPeerShardInfoV2

lwojtow:


SPEC.md
```diff
@@ -105,12 +105,23 @@ The test index makes use of symbolic language in describing connection and messa
 
 ### ZG-CONFORMANCE-004
     
-    The node responds with mtLEDGER_DATA for mtGET_LEDGER.
+    The node responds with mtLEDGER_DATA for mtGET_LEDGER with iType == LiBase.
```
niklaslong, line 108: It might be worth concisely documenting what `iType` and `LiBase` mean here?

SPEC.md
```diff
@@ -105,12 +105,23 @@ The test index makes use of symbolic language in describing connection and messa
...
     
     <>
     -> mtGET_LEDGER with iType == LiBase
     <- mtLEDGER_DATA
-    or:
 
+### ZG-CONFORMANCE-005
+
+    The node responds with mtLEDGER_DATA for mtGET_LEDGER with iType == LiAsNode.
```
niklaslong, line 116: ditto

SPEC.md
lwojtow: As per slack: agreed to elaborate on this in the near future.

SPEC.md
```diff
@@ -105,12 +105,23 @@ The test index makes use of symbolic language in describing connection and messa
...
+
+    <>
     -> mtGET_LEDGER with iType == LiAsNode
     <- mtLEDGER_DATA
+
+### ZG-CONFORMANCE-006
+
+    The node requests mtGET_PEER_SHARD_INFO_V2 after start up.
+
+    ->
```
niklaslong, line 126: Does this include the handshake or is that message sent immediately?

lwojtow: Looking at the code it seems like it's after the handshake.

niklaslong: We might want to find a way to make that clear in the symbolic language if we ever have cases that don't include the handshake, for now it's fine 👍

niklaslong (approved):

PR #33, opened 2022-08-24
Feature: add testnet

lwojtow:
1. Refactor to make node starting more configurable.
2. TestNet struct to start/stop a test net bound to localhost.


src/setup/node.rs
```diff
@@ -36,11 +37,11 @@ impl Node {
         })
     }
 
-    pub async fn start_with_peers(peers: Vec<SocketAddr>) -> Result<Self> {
-        let mut node = Self::new()?;
+    pub async fn start(path: PathBuf, ip_addr: IpAddr, peers: Vec<SocketAddr>) -> Result<Self> {
```
ljedrz, line 40: we don't need `new` and `start`, as they do the same thing now

src/setup/testnet/mod.rs
```diff
@@ -0,0 +1,199 @@
...
+            if let Err(e) = node.stop() {
+                eprintln!("Unable to stop node: {:?}", e);
+            }
+        });
+        self.cleanup().await?;
+        Ok(())
+    }
+
+    // Creates `validators.txt` file with keys of all nodes.
+    async fn build_validators_file_contents(&self) -> Result<String, fmt::Error> {
```
ljedrz, line 84:
```suggestion
    async fn create_validators_file(&self) -> Result<String, fmt::Error> {
```

src/setup/testnet/mod.rs
```diff
@@ -0,0 +1,199 @@
...
+    use_stdout: bool,
+}
+
+impl Default for TestNet {
+    fn default() -> Self {
+        Self {
+            setups: [
+                NodeSetup::new(
+                    "127.0.0.1".parse().unwrap(),
+                    "nHUSqn9qjEF7JJkVqvY7BFLMKdqP5KLLEjo5oB4QH43ADDndRawB".into(),
```
ljedrz, line 38: where do these come from? shouldn't they be some `const`s (e.g. `const validator_keys: [&str; 3] = [...]`) for better visibility?

src/setup/testnet/mod.rs
```diff
@@ -0,0 +1,199 @@
...
+        writeln!(&mut config_str, "[validators]")?;
+        for n in &self.setups {
+            writeln!(&mut config_str, "{}", n.validator_key)?;
+        }
+        Ok(config_str)
+    }
+
+    // Removes ~/.ziggurat/testnet directory
+    async fn cleanup(&self) -> io::Result<()> {
+        if let Err(e) = fs::remove_dir_all(build_testnet_path()?) {
```
ljedrz, line 95: the testnet path should be recorded and kept in the `Testnet` struct

src/setup/testnet/mod.rs
lwojtow: Could be. It's only the question whether we want stuff to be split horizontally or vertically.

src/setup/node.rs
```diff
@@ -60,7 +61,7 @@ impl Node {
         self
     }
 
-    pub async fn start(&mut self) -> Result<()> {
+    pub async fn start_process(&mut self) -> Result<()> {
```
ljedrz, line 69: we probably want this one to be private, so that we only have a single function in the tests that creates and starts nodes

src/tests/conformance/handshake.rs
```diff
@@ -11,10 +16,18 @@ async fn handshake_when_node_receives_connection() {
...
+            .join(ZIGGURAT_CONFIG),
+        IpAddr::V4(Ipv4Addr::LOCALHOST),
+        vec![],
+    )
+    .await
+    .unwrap();
 
     // Start synthetic node.
-    let synth_node = SyntheticNode::new(&TestConfig::new().unwrap()).await;
+    let synth_node = SyntheticNode::new(&TestConfig::default()).await;
```
ljedrz, line 30:
```suggestion
    let synth_node = SyntheticNode::new(&Default::default()).await;
```
applies to all such calls

src/setup/testnet/token1.txt
```diff
@@ -0,0 +1,8 @@
+eyJtYW5pZmVzdCI6IkpBQUFBQUZ4SWUyUDhJZjJTTlFvL3MzZDZReDl6Wld3cVZyWkc1N3Vh
```
ljedrz, line 1: IMO these tokens are small enough that we could keep them in the source files together with the keys, but I don't mind that much

src/setup/config.rs
```diff
@@ -2,24 +2,25 @@ use std::{
...
     path::PathBuf,
 };
 
 use anyhow::Result;
 use serde::Deserialize;
 
 // Ziggurat's configuration directory and file. Caches are written to this directory.
-const CONFIG: &str = ".ziggurat";
-const CONFIG_FILE: &str = "config.toml";
+pub const ZIGGURAT_CONFIG: &str = ".ziggurat";
```
ljedrz, line 14:
```suggestion
pub const ZIGGURAT_CONFIG_DIR: &str = ".ziggurat";
```

src/setup/config.rs
```diff
@@ -2,24 +2,25 @@ use std::{
...
+// Configuration file with paths to start rippled.
+pub const CONFIG_FILE: &str = "config.toml"; // TODO rename to something better.
```
ljedrz, line 16:
```suggestion
pub const ZIGGURAT_CONFIG: &str = "config.toml"; // TODO rename to something better.
```

src/setup/testnet/mod.rs
ljedrz: whatever `cargo fmt` prefers :stuck_out_tongue:

src/setup/node.rs
lwojtow: It seems like ideally there should be some `NodeBuilder`. As it is now, we have some functions that must be called before start, but type system doesn't prevent it :-/

lwojtow: Refactored to builder pattern.

src/setup/node.rs
```diff
@@ -26,47 +26,39 @@ pub struct Node {
...
-
     pub fn addr(&self) -> SocketAddr {
         self.config.local_addr
     }
 
-    pub fn initial_peers(&mut self, addrs: Vec<SocketAddr>) -> &mut Self {
-        self.config.initial_peers = addrs.into_iter().collect();
-        self
-    }
+    // TODO change to consume self, it's probably useless now anyway
```
ljedrz, line 33: maybe it could be included in the `Drop` impl instead? that one takes `&mut self`

src/setup/testnet/mod.rs
lwojtow: The thing is, it doesn't really create the file, that's done in `write_validators_file`. The reason for the split is that all three nodes take the same `validators.txt` file, but it's easier to have 3 files for one in each node subdir (due to how paths are treated inside Node).

src/tests/conformance/handshake.rs
lwojtow: Done

src/setup/testnet/token1.txt
lwojtow: Let's leave it as is and we'll come back to this maybe later.

src/setup/config.rs
lwojtow: Done

src/setup/testnet/mod.rs
lwojtow: Done

src/setup/config.rs
lwojtow: Done

src/setup/node.rs
lwojtow: `stop` returns `Result`. On one hand I don't want to panic in drop, on the other I don't want to swallow it there. I'm not sure what would be nicest for the future.

src/setup/config.rs
```diff
@@ -2,24 +2,25 @@ use std::{
...
     path::PathBuf,
 };
 
 use anyhow::Result;
 use serde::Deserialize;
 
 // Ziggurat's configuration directory and file. Caches are written to this directory.
-const CONFIG: &str = ".ziggurat";
-const CONFIG_FILE: &str = "config.toml";
+pub const ZIGGURAT_CONFIG_DIR: &str = ".ziggurat";
```
ljedrz, line 14:
```suggestion
pub const ZIGGURAT_DIR: &str = ".ziggurat";
```
I now noticed that cache is put there as well

src/setup/testnet/mod.rs
```diff
@@ -0,0 +1,205 @@
...
+const VALIDATOR_KEYS: [&str; 3] = [
+    "nHUSqn9qjEF7JJkVqvY7BFLMKdqP5KLLEjo5oB4QH43ADDndRawB",
+    "nHUEsvSFTf1Snr7ZUdLxjcMW6PKcMrwwXCGZBg6xb1ePG8R4C3TS",
+    "nHUuYdS49cPfRmCXPTwu7MVVFZFFmfG7y5sRttirVMhwuD7xStQp",
+];
+
+/// A struct allowing for convenient starting and stopping a small testnet.
+pub struct TestNet {
+    // Setup information for each node. Used for writing configuration.
+    setups: [NodeSetup; 3],
```
ljedrz, line 31:
```suggestion
    setups: [NodeSetup; VALIDATOR_KEYS.len()],
```

src/setup/node.rs
```diff
@@ -171,16 +141,72 @@ impl Drop for Node {
...
+    }
+
+    /// Sets initial peers for the node.
+    pub fn initial_peers(mut self, addrs: Vec<SocketAddr>) -> Self {
+        self.config.initial_peers = addrs.into_iter().collect();
+        self
+    }
+
+    /// Sets validator token to be placed in rippled.cfg.
+    /// This will make the node to be run in validator mode.
```
niklaslong, line 165: Does this mean the node _is_ a validator or does it mean it enables connections to validators?

src/setup/node.rs
```diff
@@ -171,16 +141,72 @@ impl Drop for Node {
...
     use super::*;
+    use crate::setup::config::ZIGGURAT_CONFIG_DIR;
 
     #[ignore = "convenience test to tinker with a running node for dev purposes"]
     #[tokio::test]
     async fn start_stop_node() {
-        let mut node = Node::new().unwrap();
-
-        node.log_to_stdout(true).start().await.unwrap();
+        let mut node = NodeBuilder::new(
```
niklaslong, line 199: I wrote this test when initially bootstrapping the repo, do we still need it?

src/setup/config.rs
lwojtow: Done

src/setup/node.rs
lwojtow: It turns the node into a validator. As per: https://xrpl.org/run-rippled-as-a-validator.html

ljedrz (approved):
LGTM :+1:

src/setup/testnet/mod.rs
```diff
@@ -0,0 +1,205 @@
...
+/// Validators file name.
+const VALIDATORS_FILE_NAME: &str = "validators.txt";
+
+const VALIDATOR_KEYS: [&str; 3] = [
+    "nHUSqn9qjEF7JJkVqvY7BFLMKdqP5KLLEjo5oB4QH43ADDndRawB",
+    "nHUEsvSFTf1Snr7ZUdLxjcMW6PKcMrwwXCGZBg6xb1ePG8R4C3TS",
+    "nHUuYdS49cPfRmCXPTwu7MVVFZFFmfG7y5sRttirVMhwuD7xStQp",
+];
+
+/// A struct allowing for convenient starting and stopping a small testnet.
```
niklaslong, line 28:
```suggestion
/// A struct to conveniently start and stop a small testnet.
```

src/setup/testnet/mod.rs
```diff
@@ -0,0 +1,205 @@
...
+pub struct TestNet {
+    // Setup information for each node. Used for writing configuration.
+    setups: [NodeSetup; 3],
+    // Running nodes. Used to stopping the testnet.
```
niklaslong, line 32:
```suggestion
    // Running nodes. Used to stop the testnet.
```

src/setup/node.rs
lwojtow: Not necessarily.  Should I remove it?

src/setup/testnet/mod.rs
```diff
@@ -0,0 +1,205 @@
...
+    running: Vec<Node>,
+    // Sets whether to log the node's output to Ziggurat's output stream.
+    use_stdout: bool,
+    // Path under which all nodes will be build
```
niklaslong, line 36:
```suggestion
    // Path under which all nodes will be built.
```

src/setup/testnet/mod.rs
```diff
@@ -0,0 +1,205 @@
...
+    path: PathBuf,
+}
+
+impl TestNet {
+    #[allow(dead_code)] // for some bizarre reason clippy --all-targets reports this function as unused :-/
```
niklaslong, line 41: I think it might be because it is only ever called in the test below _and_ it is private? If you make it public, clippy should no longer complain?

niklaslong (commented):
Left a few small comments, otherwise LGTM 👍

src/setup/node.rs
niklaslong: Sure, if there's no need, let's cut down on the code we need to maintain 👍

src/setup/node.rs
niklaslong: In that case I'd suggest:
```suggestion
    /// This will configure the node to run as a validator. 
```

lwojtow: Done

src/setup/testnet/mod.rs
lwojtow: Done

lwojtow: Done

src/setup/testnet/mod.rs
lwojtow: Done

src/setup/testnet/mod.rs
lwojtow: Done

src/setup/testnet/mod.rs
lwojtow: Done

src/setup/testnet/mod.rs
lwojtow: Done

niklaslong (approved):

PR #34, opened 2022-08-29
Feature: add network id config option

lwojtow:


src/tools/constants.rs
```diff
@@ -1,7 +1,17 @@
 use std::time::Duration;
 
+#[cfg(doc)]
+use crate::setup::testnet::TestNet;
+#[cfg(doc)]
+use crate::tools::inner_node::InnerNode;
+#[cfg(doc)]
+use crate::tools::synth_node::SyntheticNode;
```
niklaslong, line 8: I'd suggest using these paths in the docs themselves: e.g. `[Testnet](crate::setup::testnet::TestNet)`. `cfg_if` is another option but that adds a dependency, so I'd avoid that for these cases.

src/tools/constants.rs
```diff
@@ -1,7 +1,17 @@
...
+
 /// Timeout for [SyntheticNode] when waiting for expected message.
 pub const EXPECTED_MESSAGE_TIMEOUT: Duration = Duration::from_secs(20);
 
 /// Channel buffer bound for [InnerNode] -> [SyntheticNode] messages.
 pub const SYNTH_NODE_QUEUE_DEPTH: usize = 100;
+
+/// [TestNet]'s network id. The number here doesn't have any significance, but cannot be 0 nor 255.
```
niklaslong, line 16: ...why? 🤔

niklaslong (approved):
Left a few comments but otherwise lgtm.

src/tools/constants.rs
lwojtow: There are some semi serious answers to this: 1. Because Richard said so. 2. Because something might break. 
In all seriousness: I don't know why. I guess, I can dig into this a little deeper if it's considered to be worth my time.

niklaslong: Wouldn't say it's high prio for now but it is curious indeed.

src/tools/constants.rs
lwojtow: Done

src/tools/constants.rs
lwojtow: Will come back to this if time allows.

PR #36, opened 2022-09-01
Feature: Add stateful node

lwojtow:
A stateful node that can load ledger from saved cache and does not clean its cache when stopping.


src/setup/node.rs
```diff
@@ -62,41 +49,15 @@ impl Node {
...
-            .current_dir(&self.meta.path)
-            .args(&self.meta.start_args)
-            .stdin(Stdio::null())
-            .stdout(stdout)
-            .stderr(stderr)
-            .spawn()
-            .expect("node failed to start");
-
-        self.wait_for_start().await;
+        let process = process::start(&self.meta, self.config.log_to_stdout);
```
ljedrz, line 54: it might be a good idea for `start` to take `&self.config` as the second param, for future-proofing

src/setup/process.rs
```diff
@@ -0,0 +1,58 @@
...
+};
+
+use tokio::io::AsyncWriteExt;
+
+use crate::{setup::config::NodeMetaData, wait_until};
+
+pub const CONNECTION_TIMEOUT: Duration = Duration::from_secs(2);
+
+/// Stops the child and collects its exit message.
+pub fn stop(mut child: Child) -> Option<String> {
```
ljedrz, line 16: not a huge fan of returning a `String` here; ideally, this would be sth like
```
enum ReturnCode {
    Success,
    ErrorCode(i32),
}
```

ljedrz: also, we should be returning a `Result<ReturnCode>`, not an `Option`, and not use `ok` in the meantime; we don't want to obscure any OS errors

src/setup/mod.rs
```diff
@@ -2,4 +2,6 @@
 
 pub mod config;
 pub mod node;
+pub mod process;
```
ljedrz, line 5: not sure if we need an extra module just to start/stop the full node; also, the `process` name is easy to get mixed up with `std::process` or `tokio::process`

README.md
```diff
@@ -42,5 +42,5 @@ Follow the steps below to save an initial state that can be loaded later for oth
         "Account": "rNGknFCRBZguXcPqC63k6xTZnonSe6ZuWt",
         "Balance": "5000000000",
    ```
-5. Copy the node's files `cp -a ~/.ziggurat/testnet/1 ~/.ziggurat/ripple_test_state`
+5. Copy the node's files `cp -a ~/.ziggurat/testnet/1 ~/.ziggurat/ripple_stateful`
```
ljedrz, line 45:
```suggestion
5. Copy the node's files `cp -a ~/.ziggurat/testnet/1 ~/.ziggurat/ripple/test_state`
```

src/setup/process.rs
```diff
@@ -0,0 +1,58 @@
...
+    net::SocketAddr,
+    process::{Child, Command, Stdio},
+    time::Duration,
+};
+
+use tokio::io::AsyncWriteExt;
+
+use crate::{setup::config::NodeMetaData, wait_until};
+
+pub const CONNECTION_TIMEOUT: Duration = Duration::from_secs(2);
```
ljedrz, line 13: any reason why it's not in https://github.com/runziggurat/xrpl/blob/main/src/tools/constants.rs?

src/setup/process.rs
```diff
@@ -0,0 +1,58 @@
...
+        .args(&meta.start_args)
+        .stdin(Stdio::null())
+        .stdout(stdout)
+        .stderr(stderr)
+        .spawn()
+        .expect("node failed to start")
+}
+
+/// Waits until given socket starts accepting connections or a timeout elapses.
+pub async fn wait_for_start(addr: &SocketAddr) {
```
ljedrz, line 48:
```suggestion
pub async fn wait_for_start(addr: SocketAddr) {
```
`SocketAddr` is `Copy`

src/setup/process.rs
```diff
@@ -0,0 +1,58 @@
...
+    wait_until!(
```
ljedrz, line 49: we shouldn't use this when proper timeout tooling is available; in this case it's https://docs.rs/tokio/latest/tokio/time/fn.timeout.html

src/setup/process.rs
```diff
@@ -0,0 +1,58 @@
...
+        CONNECTION_TIMEOUT,
+        if let Ok(mut stream) = tokio::net::TcpStream::connect(addr).await {
+            stream.shutdown().await.unwrap();
+            true
```
ljedrz, line 53: this is part of the `wait_until` setup, I see; it won't be needed with `tokio::time::timeout`

src/setup/stateful.rs
```diff
@@ -0,0 +1,72 @@
...
+    process::Child,
+};
+
+use crate::setup::{config::NodeMetaData, process, process::wait_for_start};
+
+const STATEFUL_PORT: u16 = 5005;
+
+/// Stateful node. Binds to 127.0.0.1:5005, loads data and configuration from `~/.ziggurat/ripple_stateful`.
+/// Does *not* clean its database after stopping.
+pub struct StatefulNode {
```
ljedrz, line 15: why an extra object? I'd much rather just have a flag on the plain `Node`

src/setup/stateful.rs
```diff
@@ -0,0 +1,72 @@
...
+    path::PathBuf,
+    process::Child,
+};
+
+use crate::setup::{config::NodeMetaData, process, process::wait_for_start};
+
+const STATEFUL_PORT: u16 = 5005;
+
+/// Stateful node. Binds to 127.0.0.1:5005, loads data and configuration from `~/.ziggurat/ripple_stateful`.
+/// Does *not* clean its database after stopping.
```
ljedrz, line 14: you mean it doesn't delete the saved state?

src/setup/stateful.rs
```diff
@@ -0,0 +1,72 @@
...
+        tokio::time::sleep(Duration::from_secs(10)).await;
+        node.stop();
+    }
+
+    // Convenience function to build stateful node's path.
+    fn build_stateful_path() -> io::Result<PathBuf> {
+        Ok(home::home_dir()
+            .ok_or_else(|| io::Error::new(io::ErrorKind::NotFound, "couldn't find home directory"))?
+            .join(ZIGGURAT_DIR)
+            .join(STATEFUL_DIR))
```
ljedrz, line 70: we should instead copy the contents of this dir to some temporary directory, to remove the possibility of polluting the saved test state; that way the stateful nodes can still remove their _own_ state after every test

src/setup/stateful.rs
```diff
@@ -0,0 +1,72 @@
...
+
+use std::{
+    net::{Ipv4Addr, SocketAddr, SocketAddrV4},
+    path::PathBuf,
+    process::Child,
+};
+
+use crate::setup::{config::NodeMetaData, process, process::wait_for_start};
+
+const STATEFUL_PORT: u16 = 5005;
```
ljedrz, line 11: why this port (it should be documented if it's meaningful)? shouldn't it be in the consts module?

src/setup/stateful.rs
```diff
@@ -0,0 +1,72 @@
...
+pub struct StatefulNode {
+    /// The process encapsulating the running node.
+    process: Option<Child>,
+}
+
+impl StatefulNode {
+    /// Build and starts a new stateful node.
+    pub async fn new(config_path: PathBuf, log_to_stdout: bool) -> anyhow::Result<Self> {
+        let mut meta = NodeMetaData::new(config_path)?;
+        meta.start_args.append(&mut vec![
```
ljedrz, line 24: yeah, we don't want a separate `Node` setup; this should just optionally be done in the plain `Node`

src/setup/stateful.rs
```diff
@@ -0,0 +1,72 @@
...
+        }
+    }
+}
+
+#[cfg(test)]
+mod test {
+    use std::{io, path::PathBuf, time::Duration};
+
+    use crate::setup::{config::ZIGGURAT_DIR, stateful::StatefulNode};
+    const STATEFUL_DIR: &str = "ripple_stateful";
```
ljedrz, line 54: the usual comment on the const location

README.md
ljedrz: also, this should rather reference the applicable `const`

src/setup/stateful.rs
ljedrz: also, I'd name it `NODE_STATE_DIR`

src/setup/testnet/mod.rs
```diff
@@ -195,7 +195,7 @@ mod test {
 
     use crate::setup::testnet::TestNet;
 
-    #[ignore = "convenience test to tinker with a running nodes for dev purposes"]
+    #[ignore = "convenience test to tinker with running nodes for dev purposes"]
```
ljedrz, line 198:
```suggestion
    #[ignore = "used to set up a small testnet that can be used to procure node state"]
```

src/setup/testnet/mod.rs
```diff
@@ -195,7 +195,7 @@ mod test {
...
     #[tokio::test]
     async fn should_start_stop_testnet() {
```
ljedrz, line 200:
```suggestion
    async fn run_testnet() {
```
this isn't an actual test, so it shouldn't suggest it

ljedrz (changes requested):
Left a few change requests.

Cargo.toml
```diff
@@ -10,12 +10,14 @@ anyhow = "1.0"
 async-trait = "0.1"
 base64 = "0.13"
 bytes = "1"
+fs_extra = "1.2"
```
niklaslong, line 13: Not the most well-maintained codebase ever but it'll do, I don't think there's a better alternative at the moment.

src/setup/node.rs
```diff
@@ -26,25 +36,48 @@ pub struct Node {
...
+            "1".into(),
+            "--load".into(),
+        ]);
+        let process = process::start(&meta, true);
+        wait_for_start(&SocketAddr::V4(SocketAddrV4::new(
+            STATEFUL_IP,
+            JSON_RPC_PORT,
+        )))
+        .await;
+        Ok(Node {
```
niklaslong, line 59: Suggestion: leave whitespace before return values to make them quickly parsable.

src/setup/config.rs
```diff
@@ -72,7 +72,7 @@ impl NodeMetaData {
 
 /// Fields to be written to the node's configuration file.
 #[derive(Debug)]
-pub struct NodeConfig {
+pub struct NewNodeConfig {
```
niklaslong, line 75: Why change the naming here?

src/setup/node.rs
```diff
@@ -1,20 +1,30 @@
...
+            ZIGGURAT_DIR,
+        },
+        process,
+        process::wait_for_start,
+    },
+    tools::constants::{JSON_RPC_PORT, STATEFUL_DIR, STATEFUL_IP},
 };
 
-pub const CONNECTION_TIMEOUT: Duration = Duration::from_secs(2);
+pub enum NodeConfig {
```
niklaslong, line 24: ah, I see 👍

niklaslong: How about calling this `NodeSetup` instead?

src/setup/process.rs
lwojtow: Done

src/setup/process.rs
lwojtow: Done

src/setup/stateful.rs
lwojtow: Removed

src/setup/stateful.rs
lwojtow: Done

src/setup/stateful.rs
lwojtow: Moved to separate function in Node.

src/setup/stateful.rs
lwojtow: Done

src/setup/testnet/mod.rs
lwojtow: Done

src/setup/testnet/mod.rs
lwojtow: Done

Cargo.toml
lwojtow: Noted, thanks.

src/setup/node.rs
lwojtow: Done

src/setup/config.rs
lwojtow: NewNodeConfig is used to setup a 'new' nodes (in contrary to nodes that read existing ledger data).

src/setup/node.rs
lwojtow: Done

src/setup/stateful.rs
lwojtow: Moved to consts and commented.

src/setup/process.rs
lwojtow: Done

src/setup/process.rs
lwojtow: I wanted to keep related things close to each other. Moved to consts.

src/setup/mod.rs
lwojtow: Removed

README.md
lwojtow: As agreed with Niklas this change will be made in another PR.

src/setup/node.rs

src/setup/process.rs
lwojtow: Done

src/setup/config.rs
```diff
@@ -1,27 +1,29 @@
...
 use serde::Deserialize;
 
-// Ziggurat's configuration directory and file. Caches are written to this directory.
+use crate::setup::node::NodeConf;
+
+/// Ziggurat's configuration directory.
 pub const ZIGGURAT_DIR: &str = ".ziggurat";
-// Configuration file with paths to start rippled.
+
+/// Ziggurat's Ripple's subdir
```
Rqnsom, line 11: We are missing a dot at the end of the comment. It seems like that's a convention here.
(Same thing on L14)

src/tests/conformance/mod.rs
```diff
@@ -1,29 +1,26 @@
...
-        home::home_dir()
-            .expect("Can't find home directory")
-            .join(ZIGGURAT_DIR),
-        IpAddr::V4(Ipv4Addr::LOCALHOST),
-    )
-    .unwrap()
-    .build()
-    .await
-    .unwrap();
+    let temp = TempDir::new().expect("Unable to create TempDir");
```
niklaslong, line 18: I think we should abstract this into the builder if possible: we don't actually need direct access to this struct in tests.

src/tests/conformance/mod.rs
```diff
@@ -1,29 +1,26 @@
...
+    let mut node = NodeBuilder::new(None, temp.path().to_path_buf())
```
niklaslong, line 19: The path could be the same for each case by default, we're always running the cases in sequence.

src/setup/node.rs
```diff
@@ -1,193 +1,187 @@
...
     }
+}
 
-    fn generate_config_file(&self) -> Result<()> {
-        let path = self.config.path.join(RIPPLED_CONFIG);
-        let content = RippledConfigFile::generate(&self.config)?;
-        fs::write(path, content)?;
+/// Fields to be written to the node's configuration file.
+#[derive(Debug)]
+pub struct NodeConf {
```
niklaslong, line 127: Nit: I wouldn't rename this to stay consistent with the rest of the naming used throughout, `NodeConfig` and `config` are only 2 letters more.

src/setup/node.rs
```diff
@@ -1,193 +1,187 @@
...
+            network_id: None,
+        }
     }
+}
 
-    fn cleanup(&self) -> io::Result<()> {
-        self.cleanup_config_file()?;
-        self.cleanup_cache()
-    }
+pub struct Node {
```
niklaslong, line 155: The node should hold the `NodeConfig` and the `NodeMetaData` since we may need access to those fields in future test assertions.

src/setup/node.rs
```diff
@@ -1,193 +1,187 @@
...
+    addr: SocketAddr,
+}
 
-    fn cleanup_config_file(&self) -> io::Result<()> {
-        let path = self.config.path.join(RIPPLED_CONFIG);
-        match fs::remove_file(path) {
-            // File may not exist, so we suppress the error.
-            Err(e) if e.kind() != std::io::ErrorKind::NotFound => Err(e),
-            _ => Ok(()),
+impl Node {
```
niklaslong, line 160: I like the builder pattern you used, perhaps having a `builder` function on `Node` that calls the builder's `new` method will make it a little more ergonomic to work with? E.g.:
```
let node = Node::builder(state, path).set_field_smth().set_another_field();
```

src/setup/node.rs
```diff
@@ -1,193 +1,187 @@
...
+        self.addr
     }
 }
 
 impl Drop for Node {
     fn drop(&mut self) {
         // We should avoid a panic.
         if let Err(e) = self.stop() {
-            println!("Failed to stop the node: {}", e);
+            eprintln!("Failed to stop the node: {}", e);
```
niklaslong, line 184: 👍

src/tests/conformance/mod.rs
lwojtow: We don't need direct access, but `TempDir` has special `drop` impl: it removes all content.

src/setup/node.rs
lwojtow: Done

src/setup/config.rs
lwojtow: Done

src/setup/config.rs
```diff
@@ -46,6 +48,7 @@ impl NodeMetaData {
     pub fn new(config_path: PathBuf) -> Result<NodeMetaData> {
         // Read Ziggurat's configuration file.
         let path = config_path.join(ZIGGURAT_CONFIG);
+        // println!("zig: {:?}", path);
```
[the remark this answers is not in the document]
lwojtow: Done. I kind of expected that I missed something while cleaning :)

src/setup/node.rs
lwojtow: So `set_field_*` would return builder (in order to be able to chain it)? At what point would `start` be called? I don't know how to achieve above line syntactically, to be honest.

src/setup/node.rs
lwojtow: Done

src/tests/conformance/mod.rs
lwojtow: I guess this comment touches the same issue as above, so my answer still applies here:
`We don't need direct access, but TempDir has special drop impl: it removes all content.
Edit: so IOW it can be done, sure: we would need to do manual cleaning at the end of the test, preferably also at the beginning. On the other hand there would be one dependency less.`

src/tests/conformance/mod.rs
lwojtow: As agreed via slack, we're leaving it as is for now.

src/tests/conformance/mod.rs
lwojtow: As agreed via slack, we're leaving it as is for now.

niklaslong (approved):

PR #37, opened 2022-09-12
Refactor: Check for server status in stateful node test

lwojtow:


src/tools/rpc.rs
```diff
@@ -0,0 +1,66 @@
...
+    Client::new()
+        .post(JSON_RPC_ADDRESS)
+        .header(CONTENT_TYPE, "application/json")
+        .header(ACCEPT, "application/json")
+        .json(request)
+}
+
+fn build_server_info_request() -> RpcRequest {
+    RpcRequest {
+        id: "1".to_string(),
```
niklaslong, line 40:
```suggestion
        id: String::from("1"),
```

src/tools/rpc.rs
```diff
@@ -0,0 +1,66 @@
...
+        method: "server_info".to_string(),
```
niklaslong, line 41:
```suggestion
        method: String::from("server_info"),
```

src/tools/rpc.rs
```diff
@@ -0,0 +1,66 @@
...
+        api_version: 1,
```
niklaslong, line 42: We may want this to be a constant?

src/tools/rpc.rs
lwojtow: Done

src/tools/rpc.rs
lwojtow: Done

src/tools/rpc.rs
lwojtow: Done

niklaslong (approved):

PR #41, opened 2022-09-19
Feature: add check for have transactions info

lwojtow:


SPEC.md
```diff
@@ -138,5 +138,12 @@ The test index makes use of symbolic language in describing connection and messa
...
     <>
-    -> TmGetObjectByHash / OtTransactions
-    <- mtTRANSACTIONS
\ No newline at end of file
+    -> mtGET_OBJECTS with r#type == OtTransactions
+    <- mtTRANSACTIONS
+
+### ZG-CONFORMANCE-009
+
+    The node should query for transaction object after receiving TmHaveTransactions packet.
```
ljedrz, line 146:
```suggestion
    The node should query for the transaction object after receiving a TmHaveTransactions packet.
```

src/tests/conformance/query/get_by_hash.rs
```diff
@@ -74,3 +75,49 @@ async fn should_get_transaction_by_hash() {
...
+    let target = TempDir::new().expect("unable to create TempDir");
+    let mut node = build_stateful_builder(target.path().to_path_buf())
+        .expect("unable to get stateful builder")
+        .start(false)
+        .await
+        .expect("unable to start stateful node");
+
+    // Wait for correct state and account data.
+    wait_for_state("proposing".into()).await;
+    let account_data = wait_for_account_data("rNGknFCRBZguXcPqC63k6xTZnonSe6ZuWt") // TODO make constant
```
ljedrz, line 92: does this value come from the saved full node state? in that case it should rather be saved (alongside some other useful values like it) in some dedicated file during the process of producing a node's state, and loaded together with it; it can be done as a follow-up

src/tests/conformance/query/get_by_hash.rs
```diff
@@ -74,3 +75,49 @@ async fn should_get_transaction_by_hash() {
...
+    // Create stateful node.
+    let target = TempDir::new().expect("unable to create TempDir");
+    let mut node = build_stateful_builder(target.path().to_path_buf())
+        .expect("unable to get stateful builder")
+        .start(false)
+        .await
+        .expect("unable to start stateful node");
+
+    // Wait for correct state and account data.
+    wait_for_state("proposing".into()).await;
```
ljedrz, line 91: the possible states should probably be available via a dedicated `enum` instead, for added type safety; this is a nice-to-have and can be done later

src/tests/conformance/query/get_by_hash.rs
```diff
@@ -74,3 +75,49 @@ async fn should_get_transaction_by_hash() {
...
 }
+
+#[tokio::test]
+async fn should_query_for_transactions_after_have_transactions() {
+    // ZG-CONFORMANCE-009
+    // Create stateful node.
+    let target = TempDir::new().expect("unable to create TempDir");
+    let mut node = build_stateful_builder(target.path().to_path_buf())
+        .expect("unable to get stateful builder")
+        .start(false)
```
ljedrz, line 86: unrelated to this PR, but shouldn't we specify logging at the builder stage?

src/tests/conformance/query/get_by_hash.rs
```diff
@@ -74,3 +75,49 @@ async fn should_get_transaction_by_hash() {
...
+    // Wait for correct state and account data.
+    wait_for_state("proposing".into()).await;
+    let account_data = wait_for_account_data("rNGknFCRBZguXcPqC63k6xTZnonSe6ZuWt") // TODO make constant
+        .await
+        .expect("unable to get account data");
+    let tx = account_data.result.account_data.previous_transaction;
+
+    // Inform about transaction via peer protocol.
+    let mut tx_hash = [0u8; 32];
+    hex::decode_to_slice(&tx, &mut tx_hash as &mut [u8])
```
ljedrz, line 99:
```suggestion
    hex::decode_to_slice(&tx, &mut tx_hash[..])
```
haven't tested it, but this should work

src/tests/conformance/query/get_by_hash.rs
```diff
@@ -74,3 +75,49 @@ async fn should_get_transaction_by_hash() {
...
+        .expect("unable to decode transaction hash");
+    let tx_hash = tx_hash.to_vec();
```
ljedrz, line 101:
```suggestion
```

src/tests/conformance/query/get_by_hash.rs
```diff
@@ -74,3 +75,49 @@ async fn should_get_transaction_by_hash() {
...
+    let payload = Payload::TmHaveTransactions(TmHaveTransactions {
+        hashes: vec![tx_hash.clone()],
```
ljedrz, line 103:
```suggestion
        hashes: vec![tx_hash.to_vec()],
```

ljedrz (commented):
left a few comments, LGTM in general :+1:

SPEC.md
lwojtow: Done

src/tests/conformance/query/get_by_hash.rs
lwojtow: Transaction hash come from ledger, but it may be different for every setup (i.e. I have always the same, you will have always the same, but different). It can be extracted during setup and used later, but:
1. Makes setup a bit more difficult.
2. I wonder if we would have to wait for something else.

lwojtow: Added TODO.

src/tests/conformance/query/get_by_hash.rs
lwojtow: It feels "closer" to what it does here, but I'm not insisting. Added TODO.

src/tests/conformance/query/get_by_hash.rs
lwojtow: Done

src/tests/conformance/query/get_by_hash.rs
lwojtow: Done

src/tests/conformance/query/get_by_hash.rs
ljedrz: yeah, it will add an extra step to the ledger dump, but it doesn't need to complicate the node setup - each hash and other element that we need could be loaded via
```
const THING: &str = include_str!("<ledger_location>/thing");
```

src/tests/conformance/query/get_by_hash.rs
lwojtow: Added TODO

ljedrz (approved):

PR #42, opened 2022-09-19
fix: update conn timeouts

Rqnsom:
- Slower machines (Mac M2, e.g.) require a longer time for node setup.
- Small doc updates.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,2 +1,3 @@
 /target
 /Cargo.lock
+/tools/__pycache__
--- a/README.md
+++ b/README.md
@@ -18,6 +18,7 @@ The Ziggurat implementation for XRPLF's `rippled` nodes.
 5. Run tests with `cargo +stable t -- --test-threads=1`.
 
 ### Initial state
+Specific tests require an initial node state to be set up.
 Follow the steps below to save an initial state that can be loaded later for other tests.
 
 #### Preparation (needs to be done once)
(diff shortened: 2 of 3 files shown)


.gitignore
```diff
@@ -1,2 +1,5 @@
 /target
 /Cargo.lock
+/tools/__pycache__
+*.swo
+*.swp
```
ljedrz, line 5: what are those? do we expect users to always have them appear in their local repo :detective:?

README.md
```diff
@@ -47,6 +48,13 @@ Follow the steps below to save an initial state that can be loaded later for oth
         "Account": "rNGknFCRBZguXcPqC63k6xTZnonSe6ZuWt",
         "Balance": "5000000000",
    ```
-5. Copy the node's files to directory referenced by constant `pub const NODE_STATE_DIR`, currently: `cp -a ~/.ziggurat/ripple/testnet/1 ~/.ziggurat/ripple/stateful`
+5. Copy the node's files to directory referenced by constant `pub const NODE_STATE_DIR`, currently:
+   ```
+   cp -a ~/.ziggurat/ripple/testnet/1/ ~/.ziggurat/ripple/stateful;
```
Rqnsom, line 53: Here's one `\` was missing

README.md
```diff
@@ -18,6 +18,7 @@ The Ziggurat implementation for XRPLF's `rippled` nodes.
 5. Run tests with `cargo +stable t -- --test-threads=1`.
 
 ### Initial state
+Specific tests require an initial node state to be set up.
```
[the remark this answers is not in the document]
ljedrz: extra docs are always welcome :+1:

.gitignore
Rqnsom: These Vim swap files can be generated when you kill the vim session. I had around five files in the local src repo before pushing this commit.

Rqnsom: And this __pycache__ was created while I was doing the `Initial state` procedure in the readme.

ljedrz: I think user-specific temporary editor files should be out of the `.gitignore`, as we don't want to support all the possible edge cases; what about `pycache`? is it an artifact of running a testnet?

Rqnsom:
> I'd usually complain that the .gitignore and doc changes are not in their own commits, but I guess all the changes are small enough for it to be acceptable
I don't mind splitting commits into smaller ones. But this commit was really tiny.

I will split my commits into these categories in the future:
fix/feat/reafctor: <.gitignore + code> changes
doc: <just the .md changes>

.gitignore
Rqnsom: Okay, I'll move Vim swap file extensions to ~/.config/git/ignore file.

Rqnsom: Done

PR #44, opened 2022-09-21
refactor: adhere to new test naming convention

Rqnsom:
- Tests are grouped into the following categories: conformance, performance, and resistance. Each test is named after the category it belongs to, in addition to what's being tested. For example, `c001_handshake_when_node_receives_connection` is the first conformance test and tests the handshake behaviour on the receiving end. The full naming convention is: `id_part_t(subtest_no)_(message type)_(extra_test_desc)`.
- SPEC.md updated accordingly

### DIFF
--- a/SPEC.md
+++ b/SPEC.md
@@ -14,6 +14,8 @@ For load testing, "reasonable load" and "heavy load" will need to be defined.
 
 The tests can be run with `cargo test` once Ziggurat is properly configured and dependencies (node instance to be tested) are satisfied. See the [README](README.md) for details.
 
+Tests are grouped into the following categories: conformance, performance, and resistance. Each test is named after the category it belongs to, in addition to what's being tested. For example, `c001_handshake_when_node_receives_connection` is the first conformance test and tests the handshake behavior on the receiving end. The full naming convention is: `id_part_t(subtest_no)_(message type)_(extra_test_desc)`.
+
 # Types of Tests
 
 ## Conformance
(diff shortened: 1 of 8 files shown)


src/tools/constants.rs
```diff
@@ -1,7 +1,7 @@
 use std::time::Duration;
 
 /// Timeout when waiting for [Node](crate::setup::node::Node)'s start.
-pub const CONNECTION_TIMEOUT: Duration = Duration::from_secs(30);
```
Rqnsom, line 4: Sorry about this. Still, my tests sometimes fail on my Mac.

SPEC.md
```diff
@@ -104,44 +106,42 @@ The test index makes use of symbolic language in describing connection and messa
     <- pong response with the same `sequence` number
 
 ### ZG-CONFORMANCE-004
```
niklaslong, line 108: I think this case can be written more succinctly, see e.g. https://github.com/runziggurat/zcash/blob/main/SPEC.md#zg-conformance-007

SPEC.md
```diff
@@ -104,44 +106,42 @@ The test index makes use of symbolic language in describing connection and messa
...
+### ZG-CONFORMANCE-005
 
     The node requests mtGET_PEER_SHARD_INFO_V2 after connection and handshake.
 
     <>
     <- mtGET_PEER_SHARD_INFO_V2
 
-### ZG-CONFORMANCE-007
 
     The node should *NOT* request mtGET_PEER_SHARD_INFO_V2 after connection if there was no handshake.
```
niklaslong, line 131: This is different behaviour, I'd suggest leaving this as a separate case to the new 005.

Rqnsom: Sure. I initially did this split because the same message was mentioned.

src/tools/constants.rs
Rqnsom: Undone. The real fix is on this line.
https://github.com/runziggurat/xrpl/pull/44/files#diff-9fb30675c3f486d6795fcb124619aadc8a7bef02685b2d39275707e7ce9a52edR38

SPEC.md
Rqnsom: Done

niklaslong (approved):

PR #49, opened 2022-09-29
refactor: node builder can build multiple nodes

Rqnsom:
- Introducing the Stateful/Stateless NodeType option when building a node
 - Note: After the commit merges, instructions in the main readme file should be done again.


src/setup/testnet/mod.rs
```diff
@@ -11,23 +11,29 @@ use std::{
...
 
+pub fn get_validator_token(file_idx: usize) -> String {
+    match file_idx {
+        0 => include_str!("token1.txt").into(),
+        1 => include_str!("token2.txt").into(),
+        2 => include_str!("token3.txt").into(),
+        _ => panic!("token file does not exist"),
+    }
+}
+
```
Rqnsom, line 36: Maybe this could be a macro in the constants.rs.
Since node.rs is fetching this, it doesn't seem right to keep it here in testnet dir.

src/setup/node.rs
```diff
@@ -43,37 +43,69 @@ pub enum ChildExitCode {
...
-        copy(&setup_path, &target, &copy_options)?;
-        if let Some(source) = state {
-            copy(&source, &target, &copy_options)?;
-        }
-        let conf = NodeConfig::new(target);
-        let meta = NodeMetaData::new(conf.path.clone())?;
+
+        let conf = NodeConfig::new();
+        let meta = NodeMetaData::new(setup_path)?;
+
```
Rqnsom, line 66: Small note here:

Metadata shouldn't really care about the TempDir target directory.
Metadata should only have a fixed dependency towards .ziggurat/ripple/setup/config.toml.

src/setup/node.rs
```diff
@@ -43,37 +43,69 @@ pub enum ChildExitCode {
...
-    pub fn new(state: Option<PathBuf>, target: PathBuf) -> Result<Self> {
-        if !target.exists() {
-            fs::create_dir_all(&target)?;
-        }
-        let mut copy_options = CopyOptions::new();
-        copy_options.content_only = true;
-        copy_options.overwrite = true;
+    pub fn new() -> Result<Self> {
         let setup_path = build_ripple_work_path()?.join(RIPPLE_SETUP_DIR);
-        copy(&setup_path, &target, &copy_options)?;
```
Rqnsom, line 61: In the legacy code, we copied the whole setup directory, which is not necessary.
TempDir doesn't care about the config.toml file.

src/tests/conformance/query/get_by_hash.rs
```diff
@@ -85,10 +85,10 @@ async fn c008_TM_HAVE_TRANSACTIONS_query_for_transactions_after_have_transaction
     // ZG-CONFORMANCE-008
     // Create stateful node.
     let target = TempDir::new().expect("unable to create TempDir");
-    let mut node = build_stateful_builder(target.path().to_path_buf())
+    let mut node = build_stateful_builder()
         .expect("unable to get stateful builder")
         // TODO consider moving `log_to_stdout` to builder state.
-        .start(false)
+        .start(target.path().to_path_buf(), NodeType::Stateful, false)
         .await
```
lwojtow, line 92: A bit of code smell here: function has 'stateful' in its name and then we also give `NodeType::Stateful`.

Rqnsom: I agree. I wouldn't say I like the name of this function above. It just does some extra work to prepare some stuff for a stateful node. We could delete it and do those extra things in every test.

```
pub fn build_stateful_builder() -> anyhow::Result<NodeBuilder> {
    Ok(Node::builder()
        .network_id(TESTNET_NETWORK_ID)
        .validator_token(get_validator_token(0))
        .add_args(vec![
            "--valid".into(),
            "--quorum".into(),
```

src/setup/node.rs
```diff
@@ -43,37 +45,74 @@ pub enum ChildExitCode {
...
     }
 
     /// Crates [Node] according to configuration and starts its process.
-    pub async fn start(self, log_to_stdout: bool) -> Result<Node> {
-        let content = RippledConfigFile::generate(&self.conf)?;
-        let path = self.conf.path.join(RIPPLED_CONFIG);
-        fs::write(path, content)?;
+    pub async fn start(
+        &mut self,
+        target: PathBuf,
```
ljedrz, line 75:
```suggestion
        target: &Path,
```
seems to me that `PathBuf` is not needed here, so we should prefer `&Path` (just like with `&[u8]` over `Vec<u8>` or `&str` over `String`)

src/setup/node.rs
```diff
@@ -202,17 +244,16 @@ impl Drop for Node {
     }
 }
 
-pub fn build_stateful_path() -> io::Result<PathBuf> {
+fn get_stateful_node_path() -> io::Result<PathBuf> {
```
ljedrz, line 247: good call - we shouldn't use "build" for anything that doesn't employ the builder pattern :+1:

src/setup/node.rs
```diff
@@ -202,17 +244,16 @@ impl Drop for Node {
...
+    // TODO support multiple nodes
     let ziggurat_path = build_ripple_work_path()?;
-    Ok(ziggurat_path.join(NODE_STATE_DIR))
+    Ok(ziggurat_path.join(STATEFUL_NODES_DIR).join("1"))
```
ljedrz, line 250: `1` for the validator index? perhaps it should be `0` to match its token index?

src/setup/testnet/mod.rs
```diff
@@ -11,23 +11,29 @@ use std::{
...
-
 const VALIDATOR_KEYS: [&str; 3] = [
     "nHUSqn9qjEF7JJkVqvY7BFLMKdqP5KLLEjo5oB4QH43ADDndRawB",
     "nHUEsvSFTf1Snr7ZUdLxjcMW6PKcMrwwXCGZBg6xb1ePG8R4C3TS",
     "nHUuYdS49cPfRmCXPTwu7MVVFZFFmfG7y5sRttirVMhwuD7xStQp",
 ];
 
+pub fn get_validator_token(file_idx: usize) -> String {
+    match file_idx {
+        0 => include_str!("token1.txt").into(),
```
ljedrz, line 30: maybe let's zero-index the filenames to never get the indices mixed up?

ljedrz: also, maybe those should be `validator_tokenX.txt`, so it's clear those are for validators?

src/setup/node.rs
```diff
@@ -202,17 +244,16 @@ impl Drop for Node {
...
 }
 
-pub fn build_stateful_builder(target: PathBuf) -> anyhow::Result<NodeBuilder> {
-    let source = build_stateful_path()?;
-    let testnet = TestNet::new()?;
-    Ok(Node::builder(Some(source), target)
+pub fn build_stateful_builder() -> anyhow::Result<NodeBuilder> {
```
ljedrz, line 253: it's a mouthful; why not `stateful_node_builder`?

ljedrz: or, WAY better, just make it a `default` method on `NodeBuilder`

Rqnsom: Yes, that sounds much better.

src/setup/node.rs
Rqnsom: The testnet creates node directories like:
```
./ziggurat/ripple/testnet/1
./zigguratripple/testnet/2
./zigguratripple/testnet/3
```

Then these are copied to `./ziggurat/ripple/stateful/`
So we are referring here to the first copied stateful node from that folder: `./ziggurat/ripple/stateful/1/`

src/setup/testnet/mod.rs
Rqnsom: Aha, yes. Agreed.

src/setup/node.rs
```diff
@@ -148,9 +185,8 @@ pub struct NodeConfig {
 }
 
 impl NodeConfig {
-    pub fn new(path: PathBuf) -> Self {
+    pub fn new() -> Self {
```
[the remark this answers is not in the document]
Rqnsom: That's what I wanted to ask.
Clippy complained about not having a Default impl once I removed the argument.
My thinking was that, yeah, we could simply remove `new`.
Will do.

ljedrz (commented):
LGTM in general, left some comments :ok_hand:

src/setup/node.rs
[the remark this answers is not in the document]
ljedrz: I'm pretty sure even `std` only keeps the non-param `new` methods just for backwards compatibility - if there are no params and it's not `async` (at least for now), `Default` is king

ljedrz:
Also, seems that this PR needs a rebase, so that the CI will be triggered.

src/setup/node.rs
```diff
@@ -43,37 +45,90 @@ pub enum ChildExitCode {
...
-    /// Creates new [NodeBuilder]. Initial state is taken from `state` path, the node will run in `target` path.
-    pub fn new(state: Option<PathBuf>, target: PathBuf) -> Result<Self> {
-        if !target.exists() {
-            fs::create_dir_all(&target)?;
-        }
-        let mut copy_options = CopyOptions::new();
-        copy_options.content_only = true;
-        copy_options.overwrite = true;
+    /// Creates new [NodeBuilder].
+    pub fn new() -> anyhow::Result<Self> {
```
ljedrz, line 65: why not `default` REEEEE :D

ljedrz: ah, it returns a `Result`; carry on then :+1:

src/setup/node.rs
```diff
@@ -43,37 +45,90 @@ pub enum ChildExitCode {
...
-        let conf = NodeConfig::new(target);
-        let meta = NodeMetaData::new(conf.path.clone())?;
+
+        let conf = NodeConfig::default();
+        let meta = NodeMetaData::new(setup_path)?;
         Ok(Self { conf, meta })
     }
 
+    /// Creates default [NodeBuilder] which can handle stateful nodes.
+    pub fn default() -> anyhow::Result<Self> {
```
ljedrz, line 74: actually, this should be `stateful` (and the other should probably be `stateless`; also, `default` implies that `Self` is returned (not a `Result`)

src/setup/node.rs
```diff
@@ -43,37 +45,90 @@ pub enum ChildExitCode {
...
+                let validators_file_dst = target.join(VALIDATORS_FILE_NAME);
+
+                let copy_options = file::CopyOptions::new();
+                file::copy(&validators_file_src, &validators_file_dst, &copy_options)?;
+
+                self.conf.network_id = None;
+                self.conf.validator_token = None;
+                self.meta = NodeMetaData::new(setup_path)?; // Reset args
+            }
+            NodeType::Testnet => (),
```
ljedrz, line 119: just to double-check: don't we need any custom config here?

src/setup/testnet/mod.rs
```diff
@@ -11,23 +11,29 @@ use std::{
...
-
 const VALIDATOR_KEYS: [&str; 3] = [
     "nHUSqn9qjEF7JJkVqvY7BFLMKdqP5KLLEjo5oB4QH43ADDndRawB",
     "nHUEsvSFTf1Snr7ZUdLxjcMW6PKcMrwwXCGZBg6xb1ePG8R4C3TS",
     "nHUuYdS49cPfRmCXPTwu7MVVFZFFmfG7y5sRttirVMhwuD7xStQp",
 ];
 
+pub fn get_validator_token(file_idx: usize) -> String {
+    match file_idx {
+        0 => include_str!("token0.txt").into(),
```
ljedrz, line 30: I think I had a comment suggesting renaming those to `validator_tokenN.txt` :detective:

src/tools/constants.rs
```diff
@@ -17,3 +17,6 @@ pub const TESTNET_NETWORK_ID: u32 = 239048;
 
 /// The default port to start a Rippled node on.
 pub const DEFAULT_PORT: u16 = 8080;
+
+/// Validators file name.
+pub const VALIDATORS_FILE_NAME: &str = "validators.txt";
```
niklaslong, line 22: We may want to split setup constants (and have them in `setup` somewhere) with test-specific ones in future.

niklaslong (approved):

PR #51, opened 2022-09-30
chore: add log_to_stdout to node builder

Rqnsom:
- also: the RPC port is moved to const


ljedrz (approved):
Good change. Now I only wonder if the path and `NodeType` should also be moved to the builder or not :thinking:, but it can happen later if we decide so.

ljedrz:
@niklaslong any thoughts? Feel free to merge otherwise.

niklaslong (approved):

PR #52, opened 2022-09-30
feat: NodeBuilder can build multiple nodes

Rqnsom:
- NodeBuilder can build an infinite amount of stateless nodes.
- NodeBuilder can build only three stateful nodes at the moment.
- Includes tests for the NodeBuilder.


src/setup/node.rs
```diff
@@ -67,21 +72,22 @@ impl NodeBuilder {
...
         Ok(Self::stateless()
             .expect("Failed to create a node builder")
-            .network_id(TESTNET_NETWORK_ID)
-            .validator_token(get_validator_token(0))
-            .add_args(vec![
-                "--valid".into(),
-                "--quorum".into(),
-                "1".into(),
-                "--load".into(),
-            ]))
```
Rqnsom, line 84: Unfortunately, I had to remove most of these from here.
And then `add_args` became useless.

src/setup/node.rs
```diff
@@ -56,8 +57,12 @@ pub enum NodeType {
 }
 
 pub struct NodeBuilder {
+    /// Node's startup configuration
     conf: NodeConfig,
+    /// Node's process metadata read from Ziggurat configuration files
     meta: NodeMetaData,
+    /// Pool of available stateful node indices which represent stateful nodes' directories
+    stateful_nodes_pool: Vec<usize>,
```
ljedrz, line 65: unless we want to be able to pick some specific ones (e.g. 0 and 2), it would suffice for this value to just be a `usize`, no?

src/setup/node.rs
```diff
@@ -260,8 +277,118 @@ impl Drop for Node {
...
+
+#[cfg(test)]
+mod test {
+    use tempfile::TempDir;
+    use tokio::time::sleep;
+
+    use super::*;
+
+    const STATELESS_NODE_CNT: usize = 3; // Any number should work
+    const STATEFUL_NODES_CNT: usize = STATEFUL_NODES_COUNT;
```
ljedrz, line 295: do we need this const :smile:?

src/setup/node.rs
```diff
@@ -260,8 +277,118 @@ impl Drop for Node {
...
+    const TOO_MANY_STATEFUL_NODES_CNT: usize = STATEFUL_NODES_CNT + 1;
```
ljedrz, line 296: this one is not really useful too IMO; we can just add `1` in the one applicable test

src/setup/node.rs
```diff
@@ -260,8 +277,118 @@ impl Drop for Node {
...
+
+    const SLEEP: Duration = Duration::from_millis(10);
```
ljedrz, line 298:
```suggestion
    const SLEEP: Duration = Duration::from_millis(100);
```
just for extra safety

src/setup/node.rs
Rqnsom:
```
        // We always expect to get indices in increasing order when popping the vector.
        let stateful_nodes_pool = (0..STATEFUL_NODES_COUNT).rev().collect::<Vec<usize>>();
```
This above happens below.

We need to ensure that the first provided stateful node starts from directory .ziggurat/ripple/stateful/1/
since that node contains some specific ledger info. But I'm thinking now, does that matter? I guess it should work with all three. 🤔

src/setup/node.rs
Rqnsom: Yeap, we probably not

src/setup/node.rs
ljedrz: yeah, and I've seen that the pool gets `pop`ped, but one could just as well subtract from `STATEFUL_NODES_COUNT`

Rqnsom: Yeah, the tests don't seem to like nodes 1 and 2. I'm gonna check more. Maybe it has something to do with the RPC IP address.

src/setup/node.rs
```diff
@@ -90,17 +96,32 @@ impl NodeBuilder {
...
 
+        let setup_path = build_ripple_work_path()?.join(RIPPLE_SETUP_DIR);
+
         match node_type {
             NodeType::Stateful => {
-                let source = get_stateful_node_path()?;
+                let node_idx = self
+                    .stateful_nodes_pool
+                    .pop()
+                    .expect("Not enough available stateful nodes");
```
lwojtow, line 106: Perhaps `Not enough stateful nodes available` sounds better?

src/setup/node.rs
```diff
@@ -90,17 +96,32 @@ impl NodeBuilder {
...
+                let source = get_stateful_node_path(node_idx)?;
 
                 let mut copy_options = dir::CopyOptions::new();
                 copy_options.content_only = true;
                 copy_options.overwrite = true;
                 dir::copy(&source, &target, &copy_options)?;
+
+                self.conf.local_addr =
+                    SocketAddr::new(VALIDOATOR_IPS[node_idx].parse().unwrap(), DEFAULT_PORT);
```
lwojtow, line 115: Typo in `VALIDOATOR`

src/setup/node.rs
lwojtow: As this function returns `Result` maybe we shouldn't do `expect` here?

src/setup/node.rs
```diff
@@ -124,6 +146,7 @@ impl NodeBuilder {
         let node = self.start_node();
         wait_for_start(node.config.local_addr).await;
 
+        self.meta = NodeMetaData::new(setup_path)?; // Reset args
```
lwojtow, line 149: Why?

src/setup/node.rs
```diff
@@ -260,8 +277,118 @@ impl Drop for Node {
...
+                .expect("Unable to start node");
+            nodes.push(node);
+        }
+
+        sleep(SLEEP).await;
+
+        for mut node in nodes {
+            node.stop().unwrap();
+        }
+    }
```
lwojtow, line 393: It's good to have these tests, but I would suggest marking them `ignored` as they don't really test for anything, just prove  that running nodes like this is possible.

src/setup/node.rs
Rqnsom: I agree.

src/setup/node.rs
Rqnsom: Basically we need to do `meta.args.clear()` here.
Since we need to clear any previously used rippled.cfg on lines 143-144:
```
    self.meta.start_args.push("--conf".into());
    self.meta.start_args.push(rippled_cfg_path.into());
```

But we cannot simply call `clear()` because we would lose any arguments that were provided by the config.toml file:
```
        // Read the args (which includes the start command at index 0).
        let args_from = |command: &str| -> Vec<OsString> {
            command.split_whitespace().map(OsString::from).collect()
        };

        // Separate the start command from the args list.
        let mut start_args = args_from(&config_file.start_command);
        let start_command = start_args.remove(0);
```

So we need to clear the `args` that we added but keep the `args` we received from above.
The best solution I found was to recreate the `NodeMetaData` object.

src/setup/node.rs
Rqnsom: > As this function returns Result maybe we shouldn't do expect here?

Not anymore. But I applied your first suggestion since it's still applicable.

Done

src/setup/node.rs
Rqnsom: Yep, the issue was with the RPC address `127.0.0.1:5005`, and other nodes use `127.0.0.2:5005`, `127.0.0.3:5005`

Done

src/setup/node.rs
Rqnsom: Done

src/setup/node.rs
Rqnsom: Done

src/setup/node.rs
Rqnsom: So is it okay to keep them alive and then tag them as `ignored` in the next commit?

src/setup/node.rs
Rqnsom: Done

src/setup/node.rs
Rqnsom: Done

src/setup/node.rs
```diff
@@ -249,6 +266,14 @@ impl Node {
     pub fn addr(&self) -> SocketAddr {
         self.config.local_addr
     }
+
+    pub fn rpc_url(&self) -> String {
+        format!(
+            "http://{addr}:{port}",
+            addr = self.config.local_addr.ip(),
+            port = JSON_RPC_PORT
+        )
```
Rqnsom, line 275: Well. This is one solution for providing correct RPC URLs for node1 and node2.
But we need to provide the RPC URL everywhere. Not the fanciest solution.

An alternative could be:

- Move the URL RPC address into the RPC object. (rpc.rs)
- Make the RPC object part of the Node struct.
- Then use RPC calls directly on the node via RPC objects in tests like this (example for status.rs):

```
    let mut node = Node::builder()
        .start(target.path(), NodeType::Stateful)
        .await
        .expect("unable to start stateful node");
    node.rpc().wait_for_state("proposing".into()).await;

    let account_data = node.rpc().wait_for_account_data("rNGknFCRBZguXcPqC63k6xTZnonSe6ZuWt")
        .await
        .expect("unable to get account data");
    assert_eq!(account_data.result.account_data.balance, "5000000000");
```

src/setup/node.rs
lwojtow: I don't mind, sure.

src/setup/node.rs
Rqnsom: @ljedrz What do you think?

ljedrz: I don't have strong feelings one way or the other; we can always improve on this later on, if need be.

ljedrz (approved):
LGTM; @niklaslong?

src/setup/node.rs
Rqnsom: Then let's keep it like this for now.

PR #54, opened 2022-10-05
chore: centralize setup constants

Rqnsom:
- Constants belonging to the setup directory structure should be placed in setup/constants.rs
- Constants belonging to tools and tests directory structures should be placed in tools/constants.rs


ljedrz (approved):
I'm ok with this :+1:.

niklaslong (approved):

PR #55, opened 2022-10-05
chore: always serve the first stateful node first

Rqnsom:
- For some reason, stateful nodes 'node1' and 'node2' cannot reach a 'proposing.' server_state on Linux. Only 'node0' is capable of reaching the 'proposing.' state. That strange behaviour should be investigated further.

### DIFF
--- a/src/setup/node.rs
+++ b/src/setup/node.rs
@@ -58,12 +58,12 @@ pub enum NodeType {
 }
 
 pub struct NodeBuilder {
-    /// Node's startup configuration
+    /// Node's startup configuration.
     conf: NodeConfig,
-    /// Node's process metadata read from Ziggurat configuration files
+    /// Node's process metadata read from Ziggurat configuration files.
     meta: NodeMetaData,
-    /// Pool of available stateful node indices which represent stateful nodes' directories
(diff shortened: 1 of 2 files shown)


lwojtow:
I can confirm that on this branch all tests pass.

lwojtow (approved):
LGTM :+1:

src/setup/node.rs
```diff
@@ -98,12 +98,13 @@ impl NodeBuilder {
 
         match node_type {
             NodeType::Stateful => {
-                let node_idx = if self.stateful_nodes_pool != 0 {
-                    self.stateful_nodes_pool -= 1;
-                    self.stateful_nodes_pool
+                let node_idx = if self.stateful_nodes_counter < STATEFUL_NODES_COUNT {
```
ljedrz, line 101: nit; it would be shorter if we did `assert!(self.stateful_nodes_counter <= STATEFUL_NODES_COUNT);` after `self.stateful_nodes_counter += 1;`

ljedrz (approved):
LGTM with one nit

src/setup/node.rs
Rqnsom: Done

PR #56, opened 2022-10-05
Feature: Add a check for TmValidatorListCollection message.

lwojtow:
1. Moved key constants to higher level module.
2. Test for a valid TmValidatorListCollection message.

### DIFF
--- a/src/tests/conformance/cmd/validators.rs
+++ b/src/tests/conformance/cmd/validators.rs
@@ -1,0 +1,75 @@
+use serde::Deserialize;
+use tempfile::TempDir;
+
+use crate::{
+    protocol::codecs::binary::{BinaryMessage, Payload},
+    setup::node::{Node, NodeType},
+    tests::conformance::{PUBLIC_KEY_LENGTH, PUBLIC_KEY_TYPES},
+    tools::synth_node::SyntheticNode,
+};
+
+#[derive(Deserialize)]
+struct ValidatorList {
+    validators: Vec<Validator>,
+}
+
+#[derive(Deserialize)]
+struct Validator {
+    validation_public_key: String,
+    manifest: String,
+}
+
+#[tokio::test]
+#[allow(non_snake_case)]
+async fn c015_TM_VALIDATOR_LIST_COLLECTION_node_should_send_validator_list() {
+    // ZG-CONFORMANCE-015
+    let target = TempDir::new().expect("unable to create TempDir");
+    let mut node = Node::builder()
+        .start(target.path(), NodeType::Stateless)
+        .await
+        .expect("unable to start the rippled node");
+
+    // Create a synthetic node and connect it to rippled.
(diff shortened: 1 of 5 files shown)


src/tests/conformance/cmd/validators.rs
```diff
@@ -0,0 +1,75 @@
...
+#[derive(Deserialize)]
+struct ValidatorList {
+    validators: Vec<Validator>,
+}
+
+#[derive(Deserialize)]
+struct Validator {
+    validation_public_key: String,
+    manifest: String,
+}
```
ljedrz, line 75: I think I'd prefer to have the objects defined before the test(s)

src/tests/conformance/mod.rs
```diff
@@ -12,6 +12,14 @@ mod query;
 mod stateful;
 mod status;
 
+pub const PUBLIC_KEY_TYPES: &[u8] = &[
```
ljedrz, line 15: haven't we added this one elsewhere recently :detective:?

ljedrz: ah, it's just moved :+1:

src/tests/conformance/query/peer_shard_info.rs
```diff
@@ -96,7 +90,7 @@ async fn check_relay_for_key_type(key_type: u8, relays: u32) {
 
     // Create a dummy key with the specified key type.
     let mut key = vec![key_type]; // Place the key type as the first byte.
-    key.resize(33, 0x1); // Append 32 bytes serving as a dummy public key.
+    key.resize(PUBLIC_KEY_LENGTH + 1, 0x1); // Append 32 bytes serving as a dummy public key.
```
ljedrz, line 93:
```suggestion
    key.resize(PUBLIC_KEY_LENGTH, 0x1); // Append 32 bytes serving as a dummy public key.
```

ljedrz: that, or the `const` should be `32B`

ljedrz (commented):
LGTM in general, left some comments

src/tests/conformance/mod.rs
lwojtow: Yes, but now it has better name and it's moved up to be accessible in more places.

src/tests/conformance/query/peer_shard_info.rs
lwojtow: Thanks. It made one test fail, but I didn't run it after changing this :-/

src/tests/conformance/cmd/validators.rs
lwojtow: Done

src/tests/conformance/query/peer_shard_info.rs
lwojtow: Done

ljedrz (approved):
LGTM :+1:

PR #59, opened 2022-10-12
Feature: Add check for TmEndpoints messages.

lwojtow:
Check for:
1. A node sends TmEndpoints after the handshake.
2. A node does not send TmEndpoints without the handshake.

### DIFF
--- a/src/tests/conformance/cmd/endpoints.rs
+++ b/src/tests/conformance/cmd/endpoints.rs
@@ -1,0 +1,26 @@
+use crate::{
+    protocol::codecs::binary::{BinaryMessage, Payload},
+    tests::conformance::perform_response_test,
+    tools::config::TestConfig,
+};
+
+#[tokio::test]
+#[allow(non_snake_case)]
+async fn c019_TM_ENDPOINTS_node_should_send_endpoints_after_handshake() {
+    // ZG-CONFORMANCE-019
+
+    // Check for a TmEndpoints message.
+    let check = |m: &BinaryMessage| matches!(&m.payload, Payload::TmEndpoints(..));
+    perform_response_test(Default::default(), &check).await;
+}
+
+#[should_panic]
+#[tokio::test]
+#[allow(non_snake_case)]
+async fn c020_TM_ENDPOINTS_node_should_not_send_endpoints_if_no_handshake() {
+    // ZG-CONFORMANCE-020
+
+    // Check for a TmEndpoints message.
+    let response_check = |m: &BinaryMessage| matches!(&m.payload, Payload::TmEndpoints(..));
+    perform_response_test(TestConfig::default().with_handshake(false), &response_check).await;
+}
--- a/src/tests/conformance/cmd/mod.rs
+++ b/src/tests/conformance/cmd/mod.rs
@@ -1,3 +1,4 @@
+mod endpoints;
 mod manifest;
 mod squelch;
 mod validators;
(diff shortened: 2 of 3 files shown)


src/tests/conformance/cmd/endpoints.rs
```diff
@@ -0,0 +1,26 @@
...
+
+    // Check for a TmEndpoints message.
+    let check = |m: &BinaryMessage| matches!(&m.payload, Payload::TmManifests(..));
+    perform_response_test(Default::default(), &check).await;
+}
+
+#[should_panic]
+#[tokio::test]
+#[allow(non_snake_case)]
+async fn c020_TM_ENDPOINTS_node_should_not_send_endpoints_if_no_handshake() {
```
Rqnsom, line 20: More of a question.
Will the synth node receive any message if we skip the handshake procedure?
If not, maybe we could simplify and put all these `do_not_receive_if_no_handshake` tests under one test?

src/tests/conformance/cmd/mod.rs
```diff
@@ -1,3 +1,4 @@
+mod endpoints;
```
Rqnsom, line 1: I'm wondering if these messages belong to the command module.
Maybe a `broadcast` module would describe what we are testing here since `TmManifsets` and `TmEndpoints` are messages the node constantly broadcasts to us without any query. And `TmValidator` messages, I guess.

src/tests/conformance/cmd/endpoints.rs
```diff
@@ -0,0 +1,26 @@
...
+    tools::config::TestConfig,
+};
+
+#[tokio::test]
+#[allow(non_snake_case)]
+async fn c019_TM_ENDPOINTS_node_should_send_endpoints_after_handshake() {
+    // ZG-CONFORMANCE-019
+
+    // Check for a TmEndpoints message.
+    let check = |m: &BinaryMessage| matches!(&m.payload, Payload::TmManifests(..));
```
Rqnsom, line 13: Shouldn't we check `Payload::TmEndpoints(..))` here? 😀

Rqnsom (commented):
Looks good 👍

src/tests/conformance/cmd/endpoints.rs
lwojtow: I have to check this, at the moment IDK.

src/tests/conformance/cmd/mod.rs
lwojtow: Yes, I didn't like it either. Since we're testing what happens after connection/startup, I think any name like `connection`, `protocol`, `startup` will be more-or-less ok. We can move all tests like this in another PR once we decide how to call them.

Rqnsom: > We can move all tests like this in another PR once we decide how to call them.

Yes, sure!

But my view on the directory categories is based on the nature of the messages:
- When we send a query and expect a response - `query` mod (e.g. ping)
- When we send an order/command to a node (and not expect a direct answer) - `cmd` mod (e.g. squelch)
- When we do nothing but still receive some messages from a node - `maybe` broadcast mod (e.g. `TmValidation`)

src/tests/conformance/cmd/endpoints.rs
lwojtow: Will do in the next PR.

ljedrz (approved):
LGTM

src/tests/conformance/cmd/mod.rs
lwojtow: Agreed. Let's do in the next PR.

PR #60, opened 2022-10-13
Chore: Create broadcast module.

lwojtow:
1. Create broadcast module and move relevant tests there.
2. Reduce tests where no message should be received w/o handshake.


ljedrz (commented):
Since these are grouped as broadcasts, are we also planning to include tests where we check if multiple synthetic nodes receive some message, as opposed to direct messages from a full node?

We should add some documentation to the `broadcast` module explaining what kinds of tests are to reside there.

lwojtow:
> Since these are grouped as broadcasts, are we also planning to include tests where we check if multiple synthetic nodes receive some message, as opposed to direct messages from a full node?
I was thinking about some kind of 'relay' module as well, but that's different PR.

> We should add some documentation to the `broadcast` module explaining what kinds of tests are to reside there.
Will do.

src/tests/conformance/broadcast/mod.rs
```diff
@@ -0,0 +1,4 @@
+//! Contains tests for cases when a message is expected after a handshake.
```
ljedrz, line 1: are all of these messages actual broadcasts? otherwise a `post_handshake` module would be a more fitting term

ljedrz (approved):
LGTM :+1:

ljedrz:
Also, `query` tests probably fall under `post_handshake` as well, right? We might want to shuffle the modules a bit, but that can happen in a follow-up.

src/tests/conformance/broadcast/mod.rs
Rqnsom: > otherwise a post_handshake module would be a more fitting term
Aha, you are right. I think `TmValidation` and `TmProposeLedgedrs` are plain broadcast messages, we seem to receive them periodically.
And `TmManifests` and `TmEndpoints` are received just a few times after a handshake.

Rqnsom: Should we then have both the `broadcast` and the `post _handshake` modules?

ljedrz: `post_handshake` could be a parent module for all kinds of messages applicable then, including broadcasts.

Rqnsom: Okay, then something like this:

- tests
  - conformance 
    - handshake.rs
    - post_handshake
      - query/...
      - cmd/...
      - broadcast/...

Rqnsom (approved):

src/tests/conformance/broadcast/mod.rs
Rqnsom: The reorganization will be done in a later PR.

PR #66, opened 2022-10-19
Feat/test for mt validation message

lwojtow:
1. Fix message name in SPEC.
2. Add mtVALIDATION message test.
3. Simplify the TM_VALIDATOR_LIST_COLLECTION test.

### DIFF
--- a/src/tests/conformance/cmd/validators.rs
+++ b/src/tests/conformance/cmd/validators.rs
@@ -2,7 +2,7 @@ use serde::Deserialize;
 
 use crate::{
     protocol::codecs::binary::{BinaryMessage, Payload},
-    tests::conformance::{perform_response_test, PUBLIC_KEY_LENGTH, PUBLIC_KEY_TYPES},
+    tests::conformance::{perform_expected_message_test, PUBLIC_KEY_LENGTH, PUBLIC_KEY_TYPES},
 };
 
 #[derive(Deserialize)]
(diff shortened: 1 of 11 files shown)


src/tests/conformance/cmd/validators.rs
```diff
@@ -67,9 +52,5 @@ async fn c015_TM_VALIDATOR_LIST_COLLECTION_node_should_send_validator_list() {
         }
         false
     };
-    assert!(synth_node.expect_message(&check).await);
-
-    // Shutdown.
-    synth_node.shut_down().await;
-    node.stop().expect("unable to stop the rippled node");
+    perform_response_test(Default::default(), &check).await;
```
ljedrz, line 55: I'm wondering if there's a better name for this function that would not suggest that it has to be a response (as opposed to an "unsolicited" message like in this case) :thinking:

ljedrz (approved):
LGTM

src/tests/conformance/cmd/validators.rs
lwojtow: Perhaps something like `perform_expected_message_test`?

ljedrz: yeah, I think that's better; in any case, it can happen in a follow-up

lwojtow: Done

Rqnsom: `expect_message` 
or if more info
`expect_message_post_handshake` 

In any case, it would be nice to have documentation for this function in mod.rs

Rqnsom (approved):
LGTM

src/tests/conformance/cmd/validators.rs
lwojtow: I've added some comments for this function scenario.

PR #77, opened 2022-11-21
Feature: Test server identification too long

lwojtow:
* Chore: Rename `user_agent` config field to `ident`
* Feature: Add test for large 'Server' header during handshake.

### DIFF
--- a/src/tests/resistance/handshake.rs
+++ b/src/tests/resistance/handshake.rs
@@ -1,13 +1,17 @@
 use tempfile::TempDir;
+use tokio::time::sleep;
 
 use crate::{
-    setup::node::{Node, NodeType},
+    setup::{
+        constants::CONNECTION_TIMEOUT,
+        node::{Node, NodeType},
+    },
     tools::{config::TestConfig, synth_node::SyntheticNode},
 };
 
 #[allow(non_snake_case)]
 #[tokio::test]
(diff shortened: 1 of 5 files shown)


src/tests/resistance/handshake.rs
```diff
@@ -38,3 +42,42 @@ async fn r001_HANDSHAKE_reject_if_user_agent_too_long() {
     synth_node2.shut_down().await;
     node.stop().unwrap();
 }
+
+#[allow(non_snake_case)]
+#[tokio::test]
+async fn r002_HANDSHAKE_reject_if_server_too_long() {
```
Rqnsom, line 48: Maybe these two tests could be placed under the same SPEC test R001 like:
r001_t1_HANDSHAKE_xxx
r001_t2_HANDSHAKE_xxx

So we have different variants of the test: "test with wrong handshake fields"

Rqnsom: Because even the SPEC texts are pretty similar, they both revolve around doing an invalid handshake thing.

Rqnsom (approved):
LGTM 👍

src/tests/resistance/handshake.rs
lwojtow: Done

PR #80, opened 2022-11-24
refactor: rename BinaryCodec to MessageCodec

danielvladco:


src/protocol/codecs/binary.rs
```diff
@@ -75,13 +75,13 @@ pub struct BinaryMessage {
     pub payload: Payload,
```
Rqnsom, line 75: This filename should also be renamed from `binary.rs` to `message.rs`

danielvladco: done

Rqnsom (approved):
Once he returns, I want to check the naming here (the reasoning behind the original name for the binary.rs) with L.J., but if something more needs to be done, I can fix it later.

LGTM

lwojtow (approved):

PR #81, opened 2022-11-24
Feature: performance ping-pong test

asmie:
1. Added first performance test - ping-pong in zcash style.
2. Incorporated measure code (from zcash) to have possibility to create histogram.
3. Extended synthetic node to have read message with timeout functionality.


SPEC.md
```diff
@@ -304,6 +304,18 @@ The test index makes use of symbolic language in describing connection and messa
...
+## Performance
+
+### ZG-PERFORMANCE-001
+
+    The node behaves as expected under load from other peers.
+    1. Establish a node and synthetic peers.
+    2. Begin simulation.
+    3. Introspect node health and responsiveness through peers (latency, throughput). This could be done using `Ping`/`Pong` messages.
+    There can be different errors during testing: broken pipes to signal established but suddenly lost connections, 
+    InvalidData errors to indicate that connection couldn,t be established at all or timeout erros to indicate
```
danielvladco, line 316:
```suggestion
    InvalidData errors to indicate that connection couldn't be established at all or timeout errors to indicate 
```

SPEC.md
```diff
@@ -304,6 +304,18 @@ The test index makes use of symbolic language in describing connection and messa
...
+
+### ZG-PERFORMANCE-001
+
+    The node behaves as expected under load from other peers.
+    1. Establish a node and synthetic peers.
+    2. Begin simulation.
+    3. Introspect node health and responsiveness through peers (latency, throughput). This could be done using `Ping`/`Pong` messages.
+    There can be different errors during testing: broken pipes to signal established but suddenly lost connections, 
+    InvalidData errors to indicate that connection couldn,t be established at all or timeout erros to indicate 
+    connections where expected data has not been received.
```
danielvladco, line 317: I don't really understand this statement,
did you mean something like: "...connections where expecting data and it has not been received"?

src/tools/metrics/recorder.rs
```diff
@@ -0,0 +1,214 @@
...
+    CompositeKey, MetricKind,
+};
+
+// This is a false positive, `CompositeKey` does not depend on any interior mutable
+// types for its hashing implementation. Therefore, it is safe to use in our context.
+#[allow(clippy::mutable_key_type)]
+pub struct Snapshot(HashMap<CompositeKey, MetricVal>);
+
+impl Snapshot {
+    pub fn get_counter(&self, metric: &'static str) -> u64 {
```
danielvladco, line 18: I'm not a rust expert but I'm wondering why not use here `String`? and when is appropriate to use `String` vs `&'static str`

src/tools/metrics/recorder.rs
```diff
@@ -0,0 +1,214 @@
...
+        if let MetricVal::Counter(val) = *self.0.get(&key).unwrap() {
+            val
+        } else {
+            0
+        }
+    }
+
+    pub fn get_gauge(&self, metric: &'static str) -> f64 {
+        let key = CompositeKey::new(MetricKind::Gauge, Key::from_name(metric));
+        if let MetricVal::Gauge(val) = *self.0.get(&key).unwrap() {
```
danielvladco, line 29: why some methods use `*self` and others `self`?

SPEC.md
niklaslong, line 316:
```suggestion
    InvalidData errors to indicate that connection couldn't be established at all or timeout erros to indicate 
```

src/tests/performance/ping_pong.rs
```diff
@@ -0,0 +1,196 @@
...
+    let node_addr = node.addr();
+
+    for synth_count in synth_counts {
+        // setup metrics recorder
+        let test_metrics = TestMetrics::default();
+        // clear metrics and register metrics
+        metrics::register_histogram!(METRIC_LATENCY);
+
+        let mut synth_handles = Vec::with_capacity(synth_count);
+        let test_start = tokio::time::Instant::now();
```
niklaslong, line 116: Another option is the [`time`](https://time-rs.github.io/api/time/index.html) crate, if you'd like a more complete time-related lib; though this is fine too.

src/tests/performance/ping_pong.rs
```diff
@@ -0,0 +1,196 @@
...
+
+async fn simulate_peer(node_addr: SocketAddr) {
+    let config = TestConfig::default();
+    let mut synth_node = SyntheticNode::new(&config).await;
+
+    synth_node.connect(node_addr).await.unwrap();
+
+    for _ in 0..PINGS {
+        // Generate unique sequence for each ping
+        let seq = thread_rng().next_u32();
```
niklaslong, line 156: Don't think this'll make a big difference but we could reuse the rng by assigning it outside the loop.

src/tools/synth_node.rs
```diff
@@ -85,6 +86,25 @@ impl SyntheticNode {
...
+        &mut self,
+        duration: Duration,
+    ) -> Result<BinaryMessage, bool> {
+        let mut mess = None;
+        let res = timeout(duration, async {
+            mess = Some(self.recv_message().await);
+        })
+        .await
+        .is_ok();
+        if res {
```
niklaslong, line 101: Suggestion: `match` on the `.await` instead of calling `.is_ok`.

src/tools/metrics/recorder.rs
niklaslong: `String` has ownership over the contents of the string and is heap allocated, there's no need for that here. Since the metric name is a constant, hardcoding it into the executable and accessing it via a read only pointer to a string that'll be valid for the entire lifetime of the program seems appropriate?

src/tools/metrics/recorder.rs
niklaslong: [`HashMap::get`](https://doc.rust-lang.org/std/collections/hash_map/struct.HashMap.html#method.get) returns an `Option<&v>`, the `*` is dereferencing what is returned from the `unwrap` call here. I'd recommend reading [the chapter on this in the book](https://doc.rust-lang.org/book/ch15-02-deref.html).

niklaslong (commented):
Left some suggestions but overall this looks good to me 👍

src/tests/performance/ping_pong.rs
```diff
@@ -0,0 +1,196 @@
...
+                    time_taken_secs,
+                ));
+            }
+        }
+    }
+
+    node.stop().unwrap();
+
+    // Display results table
+    println!("{}", table);
```
lwojtow, line 145: I think `\r\n` should be before `{}` - if there are no errors then table's header is broken.

SPEC.md
asmie: Yes, I've changed the sentence if it was unclear.

src/tests/performance/ping_pong.rs
```diff
@@ -0,0 +1,200 @@
...
+                        seq: Some(s),
+                        ..
+                        }) if *s == seq && *r_type == PingType::PtPong as i32
+                    ) {
+                        metrics::histogram!(METRIC_LATENCY, duration_as_ms(now.elapsed()));
+                        matched = true;
+                    }
+                } // We can panic here as not receiving any message means that connection is dead.
+                Err(e) => panic!("Error receiving message - timeout?: {:?}", e),
+            }
```
Rqnsom, line 195: Shouldn't we actually `break` here instead? I see that's how it is implemented in the same Zcash test.
If we don't receive the message response, it may also mean that the rippled is running 100% and can't handle all requests on time.

Otherwise, checking `matched` in a `while` loop doesn't make much sense since it will never be set to `false` because we will always `panic` before

Rqnsom: And we are expecting one message every time we enter the loop, so not sure if we even need a `while` loop here.

asmie: The sense of checking matched in loop is that we can receive messages which may be not ping reply. So then matches! macro return us false and matched remain false and new loop iteration will begin and new message will be received.

And you're right - we can panic in case of any other error than timeout and timeout could be for us only the way to "skip" this ping - when we break we can send new ping with new seq and check if it will be handled.

Rqnsom: > The sense of checking matched in the loop is that we can receive messages which may be not ping reply

Yeah, I realized that once I read the implementation for the recv_message_timeout function below 😆

> when we break we can send new ping with new seq and check if it will be handled.

yes, let's try like that, I'm wondering if it affects the results

src/tools/synth_node.rs
```diff
@@ -85,6 +86,23 @@ impl SyntheticNode {
         }
     }
 
+    /// Reads a message from the inbound (internal) queue of the node. If there is no message
+    /// by the given time there is an error returned indicating if timeout occured.
+    pub async fn recv_message_timeout(
+        &mut self,
+        duration: Duration,
+    ) -> Result<BinaryMessage, bool> {
+        let mut mess = None;
```
Rqnsom, line 95: `mess` sounds like a mess 😃
```suggestion
        let mut msg = None;
```

src/tools/synth_node.rs
```diff
@@ -85,6 +86,23 @@ impl SyntheticNode {
         }
     }
 
+    /// Reads a message from the inbound (internal) queue of the node. If there is no message
+    /// by the given time there is an error returned indicating if timeout occured.
+    pub async fn recv_message_timeout(
```
Rqnsom, line 91: I really like the approach [zcash/algorand](https://github.com/runziggurat/zcash/blob/main/src/tools/synthetic_node.rs#L282) chose here for the same function. Could we reuse it? 🤔

Rqnsom (approved):
LGTM with a few comments, good job! 👍

src/tools/synth_node.rs
asmie: Reused it directly.

src/tests/performance/ping_pong.rs
asmie: Changes done. It hasn't affected the results. In fact, CPU load is not a case - no CPU thread is fully loaded during test. Apparently, connections are refused earlier - I'll investigate this after creating pure connection test.

src/tests/performance/ping_pong.rs
```diff
@@ -0,0 +1,208 @@
...
+    // ├─────────┼────────────┼────────────┼────────────┼────────────────┼────────────┼────────────┼────────────┼────────────┼────────────┼────────────────┼────────────┼──────────────┤
+    // │      30 │        150 │          0 │         50 │              7 │          0 │          0 │          0 │          0 │         45 │          14.60 │      15.83 │        41.52 │
+    // ├─────────┼────────────┼────────────┼────────────┼────────────────┼────────────┼────────────┼────────────┼────────────┼────────────┼────────────────┼────────────┼──────────────┤
+    // │      50 │        150 │          0 │         45 │              3 │          0 │          0 │          0 │          0 │          0 │          10.13 │      19.42 │        39.14 │
+    // ├─────────┼────────────┼────────────┼────────────┼────────────────┼────────────┼────────────┼────────────┼────────────┼────────────┼────────────────┼────────────┼──────────────┤
+    // │      75 │        150 │          0 │         57 │              6 │          0 │          0 │          0 │          0 │         41 │           8.23 │      25.28 │        36.63 │
```
Rqnsom, line 95: > let synth_counts = vec![1, 10, 15, 20, 30, 50];

We removed 75 from here, so I think we should refresh the table here?

lwojtow (approved):

src/tests/performance/ping_pong.rs
asmie: Updated.

niklaslong (approved):

PR #83, opened 2022-11-25
Chore: Configure SynthNode to use different ips

lwojtow:
Rippled seems to treat peers at the same IP somewhat strange: this test usually passed, but failed occasionally. Fixing IPs to be different seems to do the trick.


Rqnsom (approved):
LGTM

PR #84, opened 2022-11-25
Validator list

kylegranger:
This code (my very first rust code ever) basically used this rippled function  -- `buildValidatorList()` -- as a model, to see what steps are used to create and sign a manifest, as well as a validator blob.
https://github.com/XRPLF/rippled/blob/master/src/test/overlay/compression_test.cpp#L316

The verification side does the following:
- receives a manifest
- extracts the signatures
- recreates the manifest without the signatures
- appends a 4-char prefix to it, 
- gets a 32-byte digest from SHA512 (throwing away half of those 64 bytes),
- then does the signature verification, once for each of three different signature (matching each sign() call in the C++ code), also with the validator blob data.

Our keys are generated before hand, one key pair for the master key, and one for the signing key.  The master public key has been added in `validators.txt`.  The readme has been update to reflect this new content.


src/tests/conformance/post_handshake/validators.rs
```diff
@@ -55,3 +71,259 @@ async fn c015_TM_VALIDATOR_LIST_COLLECTION_node_should_send_validator_list() {
...
+            "invalid signing public key length: {}",
+            signing_public_bytes.len()
+        );
+    }
+    let signable_manifest =
+        create_signable_manifest(1, &master_public_bytes, &signing_public_bytes);
+
+    // 3. append manifest prefix
+    let mut prefixed_signable: Vec<u8> = vec![0; signable_manifest.len() + 4];
+    prefixed_signable[0..4].clone_from_slice(man_prefix.as_slice());
```
Rqnsom, line 282: Would maybe this do the trick? Then the `man_prefix` variable is not needed at all.
```suggestion
    prefixed_signable[0..4].clone_from_slice(b"MAN\x00");
```

Rqnsom (commented):
I have to go afk, but I will check more later, but here are some comments I've written so far

src/tests/conformance/post_handshake/validators.rs
```diff
@@ -55,3 +71,259 @@ async fn c015_TM_VALIDATOR_LIST_COLLECTION_node_should_send_validator_list() {
...
 }
+
+fn create_sha512_half_digest(buffer: &Vec<u8>) -> [u8; 32] {
+    let mut hasher = Sha512::new();
+    hasher.update(buffer);
+    let result = hasher.finalize();
+
+    // we return 32 bytes of 64-byte result
+    let mut signature = [0u8; 32];
+    for i in 0..32 {
```
asmie, line 82: Instead of the loop it could be possible to do this:
`signature.copy_from_slice(&result[..32]);`

src/tests/conformance/post_handshake/validators.rs
```diff
@@ -55,3 +71,259 @@ async fn c015_TM_VALIDATOR_LIST_COLLECTION_node_should_send_validator_list() {
...
+        manifest: mb,
+        blob: bb,
+        signature: sb,
+        version: 1,
+    });
+    synth_node
+        .unicast(node.addr(), payload)
+        .expect("unable to send message");
+
+    sleep(Duration::from_secs(300)).await;
```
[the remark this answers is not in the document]
kylegranger: The expectation is to (eventually) check the response that the message was accepted by rippled, and not invalid.  Lukasz and I decided to do this in a second follow up task/PR.
I changed the sleep to 5 seconds, enough time for me to confirm that the message is accepted by rippled (manually checking the console output).

src/tests/conformance/post_handshake/validators.rs
```diff
@@ -55,3 +71,255 @@ async fn c015_TM_VALIDATOR_LIST_COLLECTION_node_should_send_validator_list() {
...
+    manifest[i] = 0x73; // field code 3 for "SigningPubKey"
+    manifest[i + 1] = PUBLIC_KEY_SIZE as u8;
+    i += 2;
+    manifest[i..i + PUBLIC_KEY_SIZE].clone_from_slice(signing_pub_key.as_slice());
+    manifest
+}
+
+fn create_final_manifest(
+    sequence: u32,
+    public_key: &Vec<u8>,
```
lwojtow, line 160: Whenever you pass &Vec<T>, you could pass &[T]. This will save one pointer derefence (and I think clippy should complain about it).

src/tests/conformance/post_handshake/validators.rs
```diff
@@ -1,27 +1,43 @@
...
+    sequence: u32,
+    expiration: u32,
+    validators: Vec<Validator>,
+}
+
 #[tokio::test]
 #[allow(non_snake_case)]
 async fn c015_TM_VALIDATOR_LIST_COLLECTION_node_should_send_validator_list() {
     // ZG-CONFORMANCE-015
-
```
lwojtow, line 24: Please add this line back.

lwojtow (commented):
I realize this is your first rust code, so forgive me being picky about style. Nevertheless, this is fantastic piece of work :)

src/tests/conformance/post_handshake/validators.rs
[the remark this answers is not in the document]
Rqnsom: > The expectation is to (eventually) check the response that the message was accepted by rippled, and not invalid. Lukasz and I decided to do this in a second follow up task/PR.

aha, that makes sense. Could we leave a TODO comment then?

README.md
```diff
@@ -6,10 +6,86 @@ The Ziggurat implementation for XRPLF's `rippled` nodes.
...
+#
+# This file is UTF-8 with DOS, UNIX, or Mac style line endings.
+# Blank lines and lines starting with a '#' are ignored.
+#
+#
+#
+# [validators]
+#
+#   List of the validation public keys of nodes to always accept as validators.
+#
```
Rqnsom, line 26: General comment, I think this can be a separate PR:

These instructions are getting really huge. I would move lots of stuff here into a script, like we did in Algorand project. Check how instructions are simple here: https://github.com/runziggurat/algorand#getting-started

Alternative is:
This file could be put into [xrpl](https://github.com/runziggurat/xrpl)/setup/ directory,
and then we can have a `cp setup/validators.txt ~/.ziggurat......` command here

src/tests/conformance/post_handshake/validators.rs
```diff
@@ -55,3 +71,255 @@ async fn c015_TM_VALIDATOR_LIST_COLLECTION_node_should_send_validator_list() {
...
+    let message = Message::from_slice(&digest).unwrap();
+    let sig = engine.sign_ecdsa(&message, secret_key);
+    let sigser = sig.serialize_der();
+    let sigb64 = base64::encode(sigser);
+    base64::decode(sigb64).expect("unable to decode a blob")
+}
+
+#[tokio::test]
+#[allow(non_snake_case)]
+async fn c026_TM_VALIDATOR_LIST_send_validator_list() {
```
Rqnsom, line 224: I think you may have missed updating the SPEC.md document.

Usually every test we have in `.rs` files has a description in the [SPEC.md](https://github.com/runziggurat/xrpl/blob/main/SPEC.md) document

lwojtow: As much as I agree with keeping SPEC.md up to date, I think we can leave it until the test is finished. (Or maybe just put some 'placeholder' kind of comment there for now).

README.md
kylegranger: I liked the idea of file in `xrpl/setup`.  Did that, and updated readme accordingly.

src/tests/conformance/post_handshake/validators.rs
```diff
@@ -55,3 +71,255 @@ async fn c015_TM_VALIDATOR_LIST_COLLECTION_node_should_send_validator_list() {
...
+    serde_json::to_string(&vblob).unwrap()
+}
+
+fn create_signable_manifest(
+    sequence: u32,
+    public_key: &Vec<u8>,
+    signing_pub_key: &Vec<u8>,
+) -> Vec<u8> {
+    let size = 5 + 2 + public_key.len() + 2 + signing_pub_key.len();
+    let mut manifest: Vec<u8> = vec![0; size];
```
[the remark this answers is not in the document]
kylegranger: I've cleaned up the manifest serialization quite a bit already, but yeah, I'll go ahead and jump with `bytes`.  I think it's the right way to do this, and best to bite the bullet sooner rather than later.

kylegranger:
last of changes have been committed, including:
- serialization uses bytes::BytesMut
- all magic numbers moved to consts
- validators.txt file added to project, in setup folder
- readme updated, for copying this file to ~./ziggurat/...
- use &str over String, pass &[T] over &Vec<T>
- shorten sleep from 5 min to 5 secs
- test renamed, updated SPEC.md, added TODO for real test verification.
- use string literal for 4-char u32
- removed unused code, moved expiration creation to own function

src/tests/conformance/post_handshake/validators.rs
```diff
@@ -55,3 +85,212 @@ async fn c015_TM_VALIDATOR_LIST_COLLECTION_node_should_send_validator_list() {
...
+    let bb = bstr.as_bytes().to_vec();
+
+    // 8.  Get signature for blob using master private key
+    let blob_signature_bytes = sign_buffer(&signing_secret_key, &blob_bytes);
+
+    // 9. Setup payload, send it
+    let mstr = base64::encode(manifest);
+    let mb = mstr.as_bytes().to_vec();
+    let sstr = hex::encode_upper(blob_signature_bytes);
+    let sb = sstr.as_bytes().to_vec();
```
niklaslong, line 278: Small nit/suggestion: perhaps we could name these variables more explicitly, a few chars less isn't necessarily worth the loss in "understandable at a glance" quality?

src/tests/conformance/post_handshake/validators.rs
```diff
@@ -55,3 +85,212 @@ async fn c015_TM_VALIDATOR_LIST_COLLECTION_node_should_send_validator_list() {
...
+
+    // 4. Sign the signable manifest with master secret key, get master signature
+    let master_signature_bytes = sign_buffer(&master_secret_key, &prefixed_signable);
+
+    // 5. Sign it with signing private key, get signature
+    let signature_bytes = sign_buffer(&signing_secret_key, &prefixed_signable);
+
+    // 6. Create final manifest with sequence, public key, signing public key, master signature, signature
+    let manifest = create_final_manifest(
+        1,
```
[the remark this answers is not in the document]
Rqnsom: General improvement 2:

Just like they did in [their code](https://github.com/XRPLF/rippled/blob/master/src/ripple/protocol/impl/Sign.cpp#L25), we could handle the prefix from the `sign_buffer` function, e.g.:
```
let master_signature_bytes = sign_buffer(&hash_prefix, &master_secret_key, &manifest);
// or just call it `sign` instead of `sign_buffer` to be aligned with their code base
```


------
Then the current `create_final_manifest` function could be renamed to `sign_manifest`, and we could reuse the old manifest here (also make the `create_final_manifest` function smaller since we would only need to append `master_signature_bytes` and `signature_bytes` from that function to our already existing `manifest`):
```
sign_manifest(&mut manifest, &master_signature_bytes, &signature_bytes);
```

Rqnsom (commented):
I finally managed to get through this PR, it wasn't a straightforward one to understand,
but good work! Looks quite good, I left a few extra comments

kylegranger:
This should be pretty solid now.  Last round of changes included:
- refactored creating manifest, signing, appending prefix, signing buffers
- renamed many, many variables.
- restored text in `SPEC.md` mistakenly deleted (merge error)

src/tests/conformance/post_handshake/validators.rs
```diff
@@ -55,3 +86,196 @@ async fn c015_TM_VALIDATOR_LIST_COLLECTION_node_should_send_validator_list() {
...
+    let blob_bytes = base64::decode(&blob_b64).expect("unable to decode a blob");
+
+    // 8.  Get signature for blob using master private key
+    let signature = sign_buffer(&signing_secret_key, &blob_bytes);
+
+    // 9. Setup payload, send it
+    let manifest_b64 = base64::encode(signed_manifest);
+    let manifest_b64_bytes = manifest_b64.as_bytes().to_vec();
+    let siganture_hex = hex::encode_upper(signature);
+    let siganture_hex_bytes = siganture_hex.as_bytes().to_vec();
```
niklaslong, line 262:
```suggestion
    let signature_hex_bytes = signature_hex.as_bytes().to_vec();
```

src/tests/conformance/post_handshake/validators.rs
```diff
@@ -239,48 +230,40 @@ async fn c026_TM_VALIDATOR_LIST_send_validator_list() {
...
-        &master_public_bytes,
-        &signing_public_bytes,
-        &master_signature_bytes,
-        &signature_bytes,
-    );
+    //let mut prefixed_signable = BytesMut::with_capacity(1024);
+
+    // 3. Sign the signable manifest with master secret key, get master signature
+    let master_signature_bytes =
+        sign_buffer_with_prefix(MANIFEST_PREFIX, &master_secret_key, &manifest);
```
Rqnsom, line 239: I still think we can avoid this extra function which handles the `prefix` case

```suggestion
    let master_signature_bytes =
        sign_buffer(Some(MANIFEST_PREFIX), &master_secret_key, &manifest);
```

Then on the line below, we can reuse the function like this:
```rust
    // 8.  Get signature for blob using master private key
    let signature = sign_buffer(None, &signing_secret_key, &blob_bytes);
```

kylegranger: in my opinion.
- I think it's good to have a function that just does the signing -- I don't consider it 'extra'
- appending of a prefix is somewhat orthogonal to the signing.
- I don't like passing the idea of passing in 'None'.

src/tests/conformance/post_handshake/validators.rs
```diff
@@ -55,3 +86,193 @@ async fn c015_TM_VALIDATOR_LIST_COLLECTION_node_should_send_validator_list() {
...
+    // 4. Sign it with signing private key, get signature
+    let signature_bytes = sign_buffer_with_prefix(MANIFEST_PREFIX, &signing_secret_key, &manifest);
+
+    // 5. Create signed manifest with sequence, public key, signing public key, master signature, signature
+    let signed_manifest = sign_manifest(&manifest, &master_signature_bytes, &signature_bytes);
+
+    // 6. Create Validator blob.
+    let blob = create_validator_blob_json(&signed_manifest, master_public_hex);
+    let blob_b64 = base64::encode(&blob);
+    let blob_bytes = base64::decode(&blob_b64).expect("unable to decode a blob");
```
Rqnsom, line 250: Let's try to avoid doing a decode operation just after we encode something. We could instead copy the pre-encoded data into a separate variable before we do the encoding.
Or see my other suggestion below.

Rqnsom (commented):
Looks pretty good

src/tests/conformance/post_handshake/validators.rs
```diff
@@ -55,3 +86,193 @@ async fn c015_TM_VALIDATOR_LIST_COLLECTION_node_should_send_validator_list() {
...
+    let master_secret_bytes = hex::decode(master_secret_hex).expect("unable to decode hex");
+    let master_public_bytes = hex::decode(master_public_hex).expect("unable to decode hex");
+    let master_secret_key =
+        SecretKey::from_slice(master_secret_bytes.as_slice()).expect("unable to create secret key");
+
+    let signing_secret_hex = "00F963180681C0D1D51D1128096B8FF8668AFDC41CBDED511D12D390105EFDDC";
+    let signing_public_hex = "03859B76317C8AA64F2D253D3547831E413F2663AE2568F7A17E85B283CC8861E4";
+    let signing_secret_bytes = hex::decode(signing_secret_hex).expect("unable to decode hex");
+    let signing_public_bytes = hex::decode(signing_public_hex).expect("unable to decode hex");
+    let signing_secret_key = SecretKey::from_slice(signing_secret_bytes.as_slice())
```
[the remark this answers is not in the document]
kylegranger: done

src/tests/conformance/post_handshake/validators.rs
kylegranger: done

src/tests/conformance/post_handshake/validators.rs
```diff
@@ -55,3 +86,193 @@ async fn c015_TM_VALIDATOR_LIST_COLLECTION_node_should_send_validator_list() {
...
+    let signature_hex = hex::encode_upper(signature);
+    let signature_hex_bytes = signature_hex.as_bytes().to_vec();
+    let blob_b64_bytes = blob_b64.as_bytes().to_vec();
+
+    let payload = Payload::TmValidatorList(TmValidatorList {
+        manifest: manifest_b64_bytes,
+        blob: blob_b64_bytes,
+        signature: signature_hex_bytes,
+        version: 1,
+    });
```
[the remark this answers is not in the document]
kylegranger: done

src/tests/conformance/post_handshake/validators.rs
```diff
@@ -55,3 +86,193 @@ async fn c015_TM_VALIDATOR_LIST_COLLECTION_node_should_send_validator_list() {
...
+
+    buf
+}
+
+fn sign_manifest(manifest: &[u8], master_signature: &[u8], signature: &[u8]) -> BytesMut {
+    let mut buf = BytesMut::with_capacity(1024);
+
+    // copy manifest into buffer
+    buf.extend_from_slice(manifest);
+
```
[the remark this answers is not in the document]
kylegranger: done

Rqnsom (approved):
LGTM 🚀

PR #87, opened 2022-11-30
test: c026 for TmValidatorList, part 2

kylegranger:
This PR covers the second part of the c026.  It performs the actual verification that a correctly-formed `TmValidatorList` message was received by `rippled`.
- creates two synth nodes
- sends our TmValidatorList message from node 1, and then node 2 listens for incoming TmValidatorListCollection messages
- filters the messages, looking for one with a single validator (the other two received have about 12 each).
- when it finds our message, does asserts on the random sequence value and master public key.  
- If no valid message is found after 5 seconds -- e.g., if I didn't send the original list message, or if I sent an invalid message -- then the test would fail.  This test only confirms a positive result.

### DIFF
--- a/SPEC.md
+++ b/SPEC.md
@@ -313,7 +313,6 @@ The test index makes use of symbolic language in describing connection and messa
 
     Assert: sequence number in the validator list and public key in the validator match what was sent.
 
-
 ## Performance
 
 ### ZG-PERFORMANCE-001
--- a/SPEC.md
+++ b/SPEC.md
@@ -306,8 +306,7 @@ The test index makes use of symbolic language in describing connection and messa
 
 ### ZG-CONFORMANCE-026
 
-    The test client should send a valid mtVALIDATORLIST message, with both master and signature public keys, correctly serializing a manifest and validator blob.
-    The create is done, but the result is not yet checked.  That will be done in a future task/PR.
+    The test client should send a valid mtVALIDATORLIST message, with both master and signature public keys, correctly serializing a manifest and validator blob.  A second node verifies the correctness by receiving the validator list in an mtVALIDATORLISTCOLLECTION message and verifying the contents.
 
     <>
     <- mtVALIDATORLIST sent with master and signing public keys and a correctly serialized manifest and validator blob.
--- a/SPEC.md
+++ b/SPEC.md
@@ -306,10 +306,13 @@ The test index makes use of symbolic language in describing connection and messa
 
 ### ZG-CONFORMANCE-026
 
-    The test client should send a valid mtVALIDATORLIST message, with both master and signature public keys, correctly serializing a manifest and validator blob.  A second node verifies the correctness by receiving the validator list in an mtVALIDATORLISTCOLLECTION message and verifying the contents.
+    A synthetic node sends a mtVALIDATORLIST message with both master and signature public keys, correctly serializing a manifest and validator blob to the node. To verify the node has received the message, another synthetic node awaits a mtVALIDATORLISTCOLLECTION message from the node with the same validator blob sent by the first synthetic node in its mtVALIDATORLIST message.
 
     <>
-    <- mtVALIDATORLIST sent with master and signing public keys and a correctly serialized manifest and validator blob.
+    -> mtVALIDATORLIST with master and signing public keys and a correctly serialized manifest and validator blob.
+
+    Assert: sequence number in the validator list and public key in the validator match what was sent.
+
 
 ## Performance
 
--- a/src/tests/conformance/post_handshake/validators.rs
+++ b/src/tests/conformance/post_handshake/validators.rs
@@ -264,7 +264,7 @@ async fn c026_TM_VALIDATOR_LIST_send_validator_list() {
         let elapsed = start.elapsed();
         let millis = elapsed.as_millis();
         if millis > TIMEOUT_MILLIS {
-            assert!(false);
+            panic!("We have timed out, failing to receive valid TmValidatorListCollection message");
         }
         if let Payload::TmValidatorListCollection(validator_list_collection) = &m.payload {
             if let Some(blob_info) = validator_list_collection.blobs.first() {
--- a/src/tests/conformance/post_handshake/validators.rs
+++ b/src/tests/conformance/post_handshake/validators.rs
@@ -19,7 +19,7 @@ const ONE_YEAR: u32 = 86400 * 365;
 const JAN1_2000: u32 = 946684800;
 const RAND_SEQUENCE_NUMBER: u32 = 2022102584;
 const MANIFEST_PREFIX: &[u8] = b"MAN\x00";
-const TIMEOUT_MILLIS: Duration = Duration::from_secs(7);
+const WAIT_MSG_TIMEOUT: Duration = Duration::from_secs(5);
 
 // The master public key should be in the validators.txt file, in ~/.ziggurat/ripple/setup
 const MASTER_SECRET: &str = "8484781AE8EEB87D8A5AA38483B5CBBCCE6AD66B4185BB193DDDFAD5C1F4FC06";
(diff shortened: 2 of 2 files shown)


SPEC.md
```diff
@@ -306,8 +306,7 @@ The test index makes use of symbolic language in describing connection and messa
 
 ### ZG-CONFORMANCE-026
 
-    The test client should send a valid mtVALIDATORLIST message, with both master and signature public keys, correctly serializing a manifest and validator blob.
-    The create is done, but the result is not yet checked.  That will be done in a future task/PR.
+    The test client should send a valid mtVALIDATORLIST message, with both master and signature public keys, correctly serializing a manifest and validator blob.  A second node verifies the correctness by receiving the validator list in an mtVALIDATORLISTCOLLECTION message and verifying the contents.
```
Rqnsom, line 309: I checked other SPEC test descriptions, and there is no need to mention `the test` here,
maybe something like this (so it is more aligned with other test descriptions):
```suggestion
    A synthetic node sends a mtVALIDATORLIST message with both master and signature public keys, correctly serializing a manifest and validator blob to the node.  To verify the node has received the message, another synthetic node awaits a mtVALIDATORLISTCOLLECTION message from the node with the same validator blob sent by the first synthetic node in its mtVALIDATORLIST message. 
```

SPEC.md
```diff
@@ -306,8 +306,7 @@ The test index makes use of symbolic language in describing connection and messa
...
 
     <>
     <- mtVALIDATORLIST sent with master and signing public keys and a correctly serialized manifest and validator blob.
```
Rqnsom, line 312: We could use a short `Assert` statement here, which mentions the other synth node and how it awaits the mtVALIDATORLISTCOLLECTION message.

the `assert` statement example can be seen in c002 test above

Rqnsom (approved):
Looks pretty good, I left a few suggestions

src/tests/conformance/post_handshake/validators.rs
```diff
@@ -257,14 +255,39 @@ async fn c026_TM_VALIDATOR_LIST_send_validator_list() {
...
+                        .validation_public_key
+                        .eq(MASTER_PUBLIC));
+                    return true;
+                }
+            }
+        }
+        false
+    };
+    assert!(synth_node2.expect_message(&check).await);
 }
```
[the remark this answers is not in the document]
kylegranger: done

src/tests/conformance/post_handshake/validators.rs
```diff
@@ -257,14 +255,39 @@ async fn c026_TM_VALIDATOR_LIST_send_validator_list() {
...
         .expect("unable to send message");
 
-    // TODO: confirm result from rippled that the message was valid
-    // will be done in new PR
+    let start = Instant::now();
+    let check = |m: &BinaryMessage| {
+        let elapsed = start.elapsed();
+        let millis = elapsed.as_millis();
+        if millis > TIMEOUT_MILLIS {
+            assert!(false);
```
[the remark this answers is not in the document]
kylegranger: done

src/tests/conformance/post_handshake/validators.rs
```diff
@@ -257,14 +255,39 @@ async fn c026_TM_VALIDATOR_LIST_send_validator_list() {
...
+        if let Payload::TmValidatorListCollection(validator_list_collection) = &m.payload {
+            if let Some(blob_info) = validator_list_collection.blobs.first() {
+                let decoded_blob =
+                    base64::decode(&blob_info.blob).expect("unable to decode a blob");
 
-    sleep(Duration::from_secs(5)).await;
-    synth_node.shut_down().await;
-    node.stop().expect("unable to stop stateful node");
+                let text = String::from_utf8(decoded_blob)
+                    .expect("unable to convert decoded blob bytes to a string");
```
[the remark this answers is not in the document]
kylegranger: done

SPEC.md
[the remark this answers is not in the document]
kylegranger: done

src/tests/conformance/post_handshake/validators.rs
```diff
@@ -257,14 +255,39 @@ async fn c026_TM_VALIDATOR_LIST_send_validator_list() {
...
+                    .expect("unable to deserialize a validator list");
+
+                // Only our message has a single validator, so we skip the others
+                if validator_list.validators.len() == 1 {
+                    assert!(validator_list.sequence == RAND_SEQUENCE_NUMBER);
+                    assert!(validator_list.validators[0]
+                        .validation_public_key
+                        .eq(MASTER_PUBLIC));
+                    return true;
+                }
```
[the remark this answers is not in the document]
kylegranger: this change I chose not to do:  I like the nesting, because it's happening at many levels.  The indents aid (IMHO) understanding at a glance.  The situations where I do do what you suggest, is when there is a huge amount of code within the if block, and we're not normally in the parsing pattern.

SPEC.md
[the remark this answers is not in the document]
kylegranger: done

SPEC.md
kylegranger: done

SPEC.md
kylegranger: done

src/tests/conformance/post_handshake/validators.rs
```diff
@@ -18,7 +19,7 @@ const ONE_YEAR: u32 = 86400 * 365;
 const JAN1_2000: u32 = 946684800;
 const RAND_SEQUENCE_NUMBER: u32 = 2022102584;
 const MANIFEST_PREFIX: &[u8] = b"MAN\x00";
-const TIMEOUT_MILLIS: u128 = 5000;
+const TIMEOUT_MILLIS: Duration = Duration::from_secs(7);
```
Rqnsom, line 22: MILLIS? 🕵️‍♀️

Let's be more descriptive with file-scoped constants:
My suggestion, but feel free to change it:
```suggestion
const WAIT_MSG_TIMEOUT: Duration = Duration::from_secs(7);
```

src/tests/conformance/post_handshake/validators.rs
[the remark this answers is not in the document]
Rqnsom: Okay, but do consider this change:
```
                assert_eq!(validator_list.sequence, RAND_SEQUENCE_NUMBER);
                assert_eq!(validator_list.validators[0].validation_public_key, MASTER_PUBLIC);
```

I should have done this in another comment, sorry about that. Because using `eq()` directly feels strange... tbh

SPEC.md
```diff
@@ -306,10 +306,13 @@ The test index makes use of symbolic language in describing connection and messa
...
+    A synthetic node sends a mtVALIDATORLIST message with both master and signature public keys, correctly serializing a manifest and validator blob to the node. To verify the node has received the message, another synthetic node awaits a mtVALIDATORLISTCOLLECTION message from the node with the same validator blob sent by the first synthetic node in its mtVALIDATORLIST message.
 
     <>
-    <- mtVALIDATORLIST sent with master and signing public keys and a correctly serialized manifest and validator blob.
+    -> mtVALIDATORLIST with master and signing public keys and a correctly serialized manifest and validator blob.
+
+    Assert: sequence number in the validator list and public key in the validator match what was sent.
+
 
 ## Performance
```
Rqnsom, line 317: Let's remove an extra blank line here

```suggestion
    Assert: sequence number in the validator list and public key in the validator match what was sent.

## Performance
```

Rqnsom (approved):
LGTM with some final tiny comments

src/tests/conformance/post_handshake/validators.rs
```diff
@@ -257,14 +256,45 @@ async fn c026_TM_VALIDATOR_LIST_send_validator_list() {
...
+                        .validation_public_key
+                        .eq(MASTER_PUBLIC));
+                    return true;
+                }
+            }
+        }
+        false
+    };
+
+    timeout(TIMEOUT_MILLIS, async {
```
lwojtow, line 288: Why not just `assert!(synth_node.expect_message(&check).await);` ? 
I know it's different timeout, but shouldn't make much difference here.

lwojtow (approved):

src/tests/conformance/post_handshake/validators.rs
kylegranger: done

SPEC.md
kylegranger: done

Rqnsom (approved):
LGTM 🚀

Rqnsom (commented):
And let's squash these, git commit messages like '[test: review fixes](https://github.com/runziggurat/xrpl/pull/87/commits/6372b1324c59c40606f8b9ec58aeca7896d9ec49)' do not give much value to our git tree

PR #88, opened 2022-11-30
Feature: performance test: connection load

asmie:
This PR covers connection load tests ported from zcash. Test contains discovering how many connections are:
- accepted,
- rejected,
- terminated,
- timed out,
- has other errors.

There are three remaining things that would be fixed in one of the next PR:
1) receiver dropped messages when there are many synth peers simulated - need to investigate it further (maybe node code?);
2) under very heavy load (1k+) all connections time out - need to investigate if it's problem with our code or the rippled itself;
3) rippled max_peers setting as it not behaves as expected (https://xrpl.org/set-max-number-of-peers.html) - setting even very high number of peers does not allow to go over about 20 active connections.


src/tests/performance/connections.rs
```diff
@@ -0,0 +1,217 @@
...
+        match synth_node
+            .recv_message_timeout(Duration::from_millis(100))
+            .await
+        {
+            Ok((_, _message)) => continue, // consume every message ignoring it
+            Err(_timeout) => {
+                // check for broken connection
+                if !synth_node.is_connected(node_addr) {
+                    metrics::counter!(METRIC_TERMINATED, 1);
+                    synth_node.shut_down().await;
```
Rqnsom, line 211: nice!
this could be done in our [zcash impl](https://github.com/runziggurat/zcash/blob/main/src/tests/performance/connections.rs#L291) later as well probably

src/tests/performance/connections.rs
```diff
@@ -0,0 +1,217 @@
...
+
+            stats.timed_out = synth_count - stats.accepted - stats.rejected - stats.conn_error;
+        }
+        all_stats.push(stats);
+    }
+
+    node.stop().unwrap();
+
+    // Display results table
+    println!("\r\n{}", fmt_table(Table::new(&all_stats)));
```
Rqnsom, line 179: hmhm, could we add some assertions here? Like to expect that there will never be more than 21 accepted connections or something like that? 🤔

src/tests/performance/connections.rs
```diff
@@ -0,0 +1,217 @@
...
+    // │          50 │      50 │           20 │           30 │           19 │            0 │            0 │       9.53 │
+    // └─────────────┴─────────┴──────────────┴──────────────┴──────────────┴──────────────┴──────────────┴────────────┘
+
+    // maximum time allowed for a single iteration of the test
+    const MAX_ITER_TIME: Duration = Duration::from_secs(20);
+
+    /// maximum peers to configure node with
+    const MAX_PEERS: u16 = 50;
+
+    let synth_counts = vec![1u16, 5u16, 10u16, 20u16, 30u16, 50u16];
```
Rqnsom, line 96: to make it less noisy, only the first element can be used to define the type, so we could do this if you like:
```suggestion
    let synth_counts = vec![1u16, 5, 10, 20, 30, 50];

```

src/tests/performance/connections.rs
```diff
@@ -0,0 +1,217 @@
...
+    let config = TestConfig::default();
+    let mut synth_node = SyntheticNode::new(&config).await;
+
+    // Establish peer connection
+    let handshake_result = synth_node.connect(node_addr).await;
+    handshake_complete.send(()).await.unwrap();
+    match handshake_result {
+        Ok(stream) => {
+            metrics::counter!(METRIC_ACCEPTED, 1);
+            stream
```
Rqnsom, line 192: I don't see we are storing the `stream` anywhere. would this also work here? 
```suggestion
        Ok(_) => {
            metrics::counter!(METRIC_ACCEPTED, 1);
```

Cargo.toml
```diff
@@ -18,7 +18,7 @@ httparse = "1.7"
 openssl = "0.10"
 metrics = "0.20.0"
 metrics-util = "0.14.0"
-pea2pea = "0.40"
+pea2pea = "0.41"
```
Rqnsom, line 21: hmm, not sure if would it be a good idea to have this in a separate `uplift` solo commit (in this PR of course)

Rqnsom (approved):
LGTM 👍

src/tests/performance/connections.rs
```diff
@@ -0,0 +1,217 @@
...
+    const MAX_ITER_TIME: Duration = Duration::from_secs(20);
+
+    /// maximum peers to configure node with
+    const MAX_PEERS: u16 = 50;
+
+    let synth_counts = vec![1u16, 5u16, 10u16, 20u16, 30u16, 50u16];
+
+    let mut all_stats = Vec::new();
+
+    let target = TempDir::new().expect("Unable to create TempDir");
```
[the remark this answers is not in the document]
asmie: Changed.

src/tests/performance/connections.rs
asmie: Done.

src/tests/performance/connections.rs
asmie: Works :) Changed.

src/tests/performance/connections.rs
```diff
@@ -0,0 +1,217 @@
...
+        metrics::register_counter!(METRIC_TERMINATED);
+        metrics::register_counter!(METRIC_REJECTED);
+        metrics::register_counter!(METRIC_ERROR);
+
+        let mut synth_handles = Vec::with_capacity(synth_count as usize);
+        let mut synth_exits = Vec::with_capacity(synth_count as usize);
+        let (handshake_tx, mut handshake_rx) =
+            tokio::sync::mpsc::channel::<()>(synth_count as usize);
+
+        let test_start = Instant::now();
```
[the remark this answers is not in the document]
asmie: During https://github.com/runziggurat/xrpl/pull/81 review @niklaslong  suggested that standard time crate can be used as well. I'm a huge fan of using standard solutions without any external dependencies where it's possible but of course, this is not the case as we've already used tokio. I'm not so deep in the optimization issues and therefore I'm not really aware which one implementation is better (tokio or time).

src/tests/performance/connections.rs
asmie: Yes, there should be.
I'll add this in the next PR (let's call those assertions business logic). I wouldn't do this right now because after trials with the source IP logic I was not sure how the rippled should really behave when getting connections from the same or different IP addresses.

Cargo.toml
asmie: Well, this test will not work very well without 0.41 version. However I can put new PR when there will be 0.42 version (should be today) and it will only contain that one line to bump pea2pea from 0.40 to 0.42. 
So by now, I'm rolling back to 0.40.

src/tests/performance/connections.rs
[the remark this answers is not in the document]
lwojtow: I'm ok with both.

SPEC.md
```diff
@@ -324,6 +324,13 @@ The test index makes use of symbolic language in describing connection and messa
     InvalidData errors to indicate that connection couldn't be established at all or timeout errors to indicate
     that data has not been received in timely manner.
 
+### ZG-PERFORMANCE-002
+
+    The node sheds or rejects connections when necessary.
```
lwojtow, line 329: This makes it sound as a resistance test.

src/tests/performance/connections.rs
```diff
@@ -0,0 +1,216 @@
...
+                tokio::select! {
+                    _ = exit_rx => {},
+                    _ = simulate_peer(node_addr, synth_handshaken) => {},
+                };
+            }));
+        }
+
+        // Wait for all peers to indicate that they've completed the handshake portion
+        // or the iteration timeout is exceeded.
+        let _ = tokio::time::timeout(MAX_ITER_TIME, async move {
```
lwojtow, line 142: From the comment it sounds as it could be implemented with tokio's Barrier?

src/tests/performance/connections.rs
```diff
@@ -0,0 +1,216 @@
...
+            stats.terminated = snapshot.get_counter(METRIC_TERMINATED) as u16;
+            stats.rejected = snapshot.get_counter(METRIC_REJECTED) as u16;
+            stats.conn_error = snapshot.get_counter(METRIC_ERROR) as u16;
+
+            stats.timed_out = synth_count - stats.accepted - stats.rejected - stats.conn_error;
+        }
+        all_stats.push(stats);
+    }
+
+    node.stop().unwrap();
```
lwojtow, line 176: Can we use `expect` here for better error message just in case?

src/tests/performance/connections.rs
```diff
@@ -0,0 +1,216 @@
...
+        }
+    };
+
+    // Keep connection alive by consuming messages
+    loop {
+        match synth_node
+            .recv_message_timeout(Duration::from_millis(100))
+            .await
+        {
+            Ok((_, _message)) => continue, // consume every message ignoring it
```
lwojtow, line 205: Small nitpick: if we aren't using _message (and others) maybe we can just call `_`?

lwojtow (approved):

src/tests/performance/connections.rs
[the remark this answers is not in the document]
Rqnsom: hmhm, I think @niklaslong  was talking about time-rs crate:
https://time-rs.github.io/api/time/struct.Instant.html
vs
https://doc.rust-lang.org/std/time/struct.Instant.html

My general thinking here is: in async applications, we should use async versions of the API whenever possible.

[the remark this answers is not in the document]
niklaslong: Synchronous is fine too, especially if there's no specific or underlying need for async.

SPEC.md
asmie: Agree, this test could be resistance but as @Rqnsom  said, it has been classified to performance before. IMHO it should be left in this category to be consistent with other projects.

src/tests/performance/connections.rs
asmie: Done.

src/tests/performance/connections.rs
asmie: Done.

src/tests/performance/connections.rs
asmie: Done.

src/tests/performance/connections.rs
asmie: The idea here is to wait until all synth_nodes complete handshakes or timeout occur but without blocking the threads. Each thread when complete handshake is receiving messages sent by the rippled just to ensure we won't suffer any overflows due to incoming messages. Barrier would block also block thread till all other threads complete their handshakes.
So to achieve this there is tokio::mpsc where synth_node sends empty value and main thread just wait to receive anything from each thread.

Cargo.toml
asmie: Sorry, due to couple merges rebasing now commit to the first place is somehow problematic. I'll keep in mind our talk and try to produce more linear PR next time :)

src/tests/performance/connections.rs
```diff
@@ -0,0 +1,218 @@
...
+            stats.conn_error = snapshot.get_counter(METRIC_ERROR) as u16;
+
+            stats.timed_out = synth_count - stats.accepted - stats.rejected - stats.conn_error;
+        }
+        all_stats.push(stats);
+    }
+
+    // TODO: assertions with expected test results.
+
+    node.stop().expect("failed to stop node");
```
Rqnsom, line 178: just for consistency with other files
```suggestion
    node.stop().expect("unable to stop the node");
```

Rqnsom (approved):
LGTM 🚀

PR #89, opened 2022-11-30
tests: added tool for generating dummy devices and Rust array with ips

asmie:
Tool that generate dummy devices, set them hardware addresses, ip addresses and set link up. Along with that it generates Rust file with all valid IP addresses that has been added to devices.

Script currently works on Linux only and needs to be run with sudo (as there are needed privilages to create devices).

Dummy kernel module needed (should be present on every distro with default kernel).

### DIFF
--- a/tools/ips.py
+++ b/tools/ips.py
@@ -1,0 +1,51 @@
+import argparse
+import ipaddress
+import os
+import random
+import sys
+
+
+def generate_hwaddr():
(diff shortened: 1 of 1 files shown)


Rqnsom (commented):
I don't know python, so I can't provide much feedback here 🤷

The only thing I can mention is we could add the same description maybe on top of the file as documentation for the file.

PR #90, opened 2022-12-02
refactor: ping performance test can use different source IPs

asmie:
This PR gives ability for performance ping test to use different IP addresses.

Currently it's recommended to generate addresses to cover all iterations of all tests, so it would be good to generate whole 24bit subnet. When pea2pea will allow to set REUSE_ADDR option "the hack" can be removed.

### DIFF
--- a/src/tests/performance/ping_pong.rs
+++ b/src/tests/performance/ping_pong.rs
@@ -2,7 +2,6 @@ use std::{
     io::ErrorKind,
     net::{IpAddr, Ipv4Addr, SocketAddr},
     str::FromStr,
-    thread,
     time::{Duration, Instant},
 };
 
@@ -26,6 +25,7 @@ use crate::{
     },
 };
 
+const MAX_PEERS: usize = 100;
 const PINGS: u16 = 1000;
 const METRIC_LATENCY: &str = "ping_perf_latency";
 const CONNECTION_PORT: u16 = 31337;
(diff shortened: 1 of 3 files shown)


src/tests/performance/ping_pong.rs
```diff
@@ -110,10 +107,12 @@ async fn p001_t1_PING_PONG_throughput() {
 
         let mut synth_handles = Vec::with_capacity(synth_count);
         let test_start = tokio::time::Instant::now();
-        for _ in 0..synth_count {
-            synth_handles.push(tokio::spawn(simulate_peer(node_addr)));
+        for i in 0..synth_count {
+            synth_handles.push(tokio::spawn(simulate_peer(node_addr, i + used)));
         }
 
+        used += synth_count;
```
Rqnsom, line 114: would this be more efficient:
```suggestion
        for i in 0..synth_count {
            synth_handles.push(tokio::spawn(simulate_peer(node_addr, used)));
            used += 1;
        }
```

src/tests/performance/ping_pong.rs
```diff
@@ -55,53 +57,48 @@ async fn p001_t1_PING_PONG_throughput() {
...
         .unwrap();
     let node_addr = node.addr();
 
+    // This is "the hack" but is needed to perform next tests if IPS table is not empty. The
+    // standard TIME_WAIT is 60s before we can use the same addr and port again.
+    // So we're taking already used IPs and each thread in each iteration will use different IP.
+    // If the table is empty or too small, the thread itself will notice it and will use the
+    // local IP.
+    // It can be removed once pea2pea will offer REUSE_ADDR options.
+    let mut used = 0;
```
Rqnsom, line 100: maybe we could improve naming here
```suggestion
    let mut ip_idx = 0;  or // used_ip_cnt
```

lwojtow (approved):

src/tools/ips.rs
```diff
@@ -0,0 +1,8 @@
+// If the IPS array is empty then it means that source IP addresses have not been
+// generated and assigned to dummy devices. In such case tests should not bound to any
+// specific address and use local pool instead.
+
+/// Reference to a static array of IP addresses (represented as str).
+/// If array is empty generate new addresses using eg.:
+/// python3 tools/ips.py --subnet 1.1.1.0/28 --file src/tools/ips.rs --dev_prefix test_dev
+pub const IPS: &[&str] = &[];
```
Rqnsom, line 8: I can't do this on Mac, what's the size of this file after we do this?

Rqnsom (approved):
LGTM

src/tools/ips.rs
Rqnsom: Would it be possible to use `ifconfig` instead here so it also works on Mac?

src/tests/performance/ping_pong.rs
asmie: Done.

src/tests/performance/ping_pong.rs
asmie: Done.

src/tools/ips.rs
asmie: Yes, it'd be possible to use ifconfig on Mac (although you can use brew to install ip -> iproute2mac), but either one does not allow to create dummy devices.
Probably for Macs there will be a need to either create address aliases for single interface or add other type of interfaces (maybe tap and bridge it).

PR #91, opened 2022-12-02
Refactor: connection load test with different IP sources

asmie:
Connection load test now can use different source IPs. Ported assertions from zcash to check test results.

Seems that rippled still does not behave like it should accepting more connections than max_peers and terminating some other connections without a clear reason.

Currently it's recommended to generate addresses to cover all iterations of all tests, so it would be good to generate whole 24bit subnet. When pea2pea will allow to set REUSE_ADDR option "the hack" can be removed.


lwojtow:
Approved, but I guess this will be changed anyway when Lukasz makes his changes to pea2pea.

lwojtow (approved):

src/tests/performance/connections.rs
```diff
@@ -173,16 +246,48 @@ async fn p002_connections_load() {
...
+
+        // And no connection timeouts or errors
+        assert_eq!(stats.timed_out, 0, "Stats: {:?}", stats);
+        assert_eq!(stats.conn_error, 0, "Stats: {:?}", stats);
+    }
 }
 
-async fn simulate_peer(node_addr: SocketAddr, handshake_complete: Sender<()>) {
-    let config = TestConfig::default();
+async fn simulate_peer(node_addr: SocketAddr, handshake_complete: Sender<()>, thread_num: usize) {
```
Rqnsom, line 277: we could also rename `thread_num` to something else here

Rqnsom (approved):
LGTM

src/tests/performance/connections.rs
asmie: Changed.

PR #92, opened 2022-12-05
chore: update pea2pea dependency

lwojtow:
Updating pea2pea dependency with necessary refactoring.


src/tools/config.rs
```diff
@@ -48,6 +48,8 @@ pub struct SyntheticNodeTestConfig {
     pub handshake_bit_flip_shared_val: bool,
     /// Will flip random bit in a random byte of the public key.
     pub handshake_bit_flip_pub_key: bool,
+    /// Optional bound address when initiating a connection.
+    pub bound_addr: Option<SocketAddr>,
```
ljedrz, line 52: this shouldn't be node-level, but connection-level instead

src/tools/inner_node.rs
```diff
@@ -90,9 +91,13 @@ impl InnerNode {
         connector.set_verify(SslVerifyMode::NONE); // we might remove it once the keypair is solid
         let connector = connector.build();
 
+        let node = Node::new(config.pea2pea_config.clone());
+        node.start_listening()
+            .await
+            .expect("unable to start listening");
```
ljedrz, line 97:
```suggestion
```
this should happen after the protocols are enabled, even if it needs to be explicit in every test

ljedrz (changes requested):
Left a few comments.

src/tools/inner_node.rs
```diff
@@ -118,8 +122,24 @@ impl InnerNode {
...
     /// Connects to the target address.
     pub async fn connect(&self, target: SocketAddr) -> io::Result<()> {
-        self.node.connect(target).await?;
-        Ok(())
+        self.node.connect(target).await
+    }
+
+    /// Connects to the target address using outgoing_addr as source address/port.
+    /// Sets `reuse_addr` and `reuse_port` on the outgoing socket.
+    pub async fn connect_with_outgoing_addr(
```
Rqnsom, line 130: just maybe? Not sure
```suggestion
    pub async fn connect_from(
```

src/tools/inner_node.rs
```diff
@@ -118,8 +122,24 @@ impl InnerNode {
...
+        target: SocketAddr,
+    ) -> io::Result<()> {
+        let socket = match target {
+            SocketAddr::V4(_) => TcpSocket::new_v4()?,
+            SocketAddr::V6(_) => TcpSocket::new_v6()?,
+        };
+        socket.bind(outgoing_addr)?;
+        socket.set_reuseaddr(true)?;
+        socket.set_reuseport(true)?;
+        self.node.connect_using_socket(target, socket).await
```
Rqnsom, line 142: If we put this here, we cannot prepare the socket before starting the critical section of the performance test.
Maybe we can leave it out and give the user more control here just for that case above. Otherwise, we'll suffer a performance hit.

asmie (approved):
LGTM after incorporating @ljedrz  and @Rqnsom  suggestions about moving reuse*() and listening back to the tests.

PR #93, opened 2022-12-05
Feature: IP address generator can add addresses to devs and work on Mac 

asmie:
Now there is ability to either generate new dummy devs and set them addresses or add addresses to existing device.
After testing new option --op remove was added to ensure user can easily delete all used devs and addresses from system.
Adding address to existing dev works well on Mac (like adding to lo0).

### DIFF
--- a/tools/ips.py
+++ b/tools/ips.py
@@ -69,7 +69,7 @@ parser.add_argument('--subnet', nargs='?', default='1.1.1.0/28', help='Subnet to
 parser.add_argument('--file', nargs='?', default='ips.rs', help='Output file with IPs (default ips.rs)', type=str)
 parser.add_argument('--op', nargs='?', default='add', help='add/remove operation', type=str)
 parser.add_argument('--dev_prefix', nargs='?', help='How to prefix dummy devs (Linux only)(def: test_dummyX)', type=str)
-parser.add_argument('--dev', nargs='?', help='Device to add addresses (Linux and Mac)(default: lo0)', type=str)
+parser.add_argument('--dev', nargs='?', help='Device to add addresses (Linux and Mac)', type=str)
 
 args = parser.parse_args()
 
@@ -101,9 +101,6 @@ else:
 
 if args.dev is not None:
     dev = args.dev
-else:
-    if args.dev_prefix is None:
-        dev = 'lo0'
 
 if sys.platform.startswith('linux'):
     if args.dev_prefix is not None :


tools/ips.py
```diff
@@ -29,32 +43,89 @@ def generate_dev(device_name, ip_addr):
...
+    operation_addr = 'add_addr_to_existing_dev'
+else:
+    operation_dev = 'remove_dev'
+    operation_addr = 'delete_addr_from_existing_dev'
+
+if args.dev is not None:
+    dev = args.dev
+else:
+    if args.dev_prefix is None:
+        dev = 'lo0'
```
Rqnsom, line 106: I don't know python, but wouldn't would this work here?
```suggestion
else if args.dev_prefix is None:
        dev = 'lo0'
```

Rqnsom: But now that I see this code above:
```py
if args.dev_prefix is None and args.dev is None:
    print('Either --dev_prefix or --dev must be specified')
    sys.exit(0)
```
Seems like we should never be able to enter this case.
I think we can simply use this instead:
```py
if args.dev is not None:
    dev = args.dev
```

Rqnsom (approved):
LGTM,
a few small comments (I should learn python)

README.md
```diff
@@ -16,7 +16,21 @@ cp setup/validators.txt ~/.ziggurat/ripple/setup
...
+   ```
+   Under MacOS (to add whole subnet to loopback device):
+   ```
+   sudo python3 ./tools/ips.py --subnet 1.1.1.0/24 --file src/tools/ips.rs --dev lo0
+   ```
+   Adding new address to existing device works under Linux too. You can adjust subnet and device name as you wish. Read ./tools/ips.py for more details or run:
+   ```
+   python3 ./tools/ips.py --help
+   ```
+   for options.
```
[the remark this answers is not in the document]
asmie: Changed.

README.md
```diff
@@ -16,7 +16,21 @@ cp setup/validators.txt ~/.ziggurat/ripple/setup
     path = "<path to the directory where you built rippled>"
     start_command = "./rippled"
     ```
-5. Run tests with `cargo +stable t -- --test-threads=1`.
+5. Create package of IP addresses to be used as source IP for test nodes. Those source IP addresses are important for performance tests. Beeing in the repository directory run eg.:
+   Under Linux (to generate dummy devices with addresses):
```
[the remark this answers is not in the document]
asmie: Changed.

tools/ips.py
asmie: Removed.

README.md
```diff
@@ -16,7 +16,17 @@ cp setup/validators.txt ~/.ziggurat/ripple/setup
     path = "<path to the directory where you built rippled>"
     start_command = "./rippled"
     ```
-5. Run tests with `cargo +stable t -- --test-threads=1`.
+5. Create a package of IP addresses which are required for performance tests. From the root repository directory run, e.g.:
+   Under Linux (to generate dummy devices with addresses):
+   ```
+   sudo python3 ./tools/ips.py --subnet 1.1.1.0/24 --file src/tools/ips.rs --dev_prefix test_zeth
+   ```
+   Under MacOS or Linux (to add whole subnet to loopback device - under Linux: lo, MacOS: lo0):
```
Rqnsom, line 24: oh, right, we can probably remove the below part:
```
##### Mac users
Make sure these two `127.0.0.x` (where `x != 1`) addresses are enabled:

    sudo ifconfig lo0 alias 127.0.0.2 up;
    sudo ifconfig lo0 alias 127.0.0.3 up;
```
On line 39. I cannot leave a comment there...

Rqnsom (approved):

README.md
asmie: Not yet, VALIDATOR_IPS are hardcoded local IPs (three addresses) used in some other places (like c024). Currently it's demanded to be exactly those two additional addresses and they cannot be from any other subnet (that could  be changed by refactoring the tests). 
So in this PR I'd leave it as is.

PR #94, opened 2022-12-06
chore: pea2pea uplift and test adjustments

asmie:
Pea2pea uplifted to 0.43.
Connection and ping test updated to create sockets outside of each thread (socket pool) and then use it.


src/tools/inner_node.rs
```diff
@@ -122,6 +122,12 @@ impl InnerNode {
         Ok(())
     }
 
+    /// Connects to the target address.
+    pub async fn connect_from(&self, target: SocketAddr, socket: TcpSocket) -> io::Result<()> {
+        self.node.connect_using_socket(target, socket).await?;
+        Ok(())
+    }
```
Rqnsom, line 129:
```suggestion
    pub async fn connect_from(&self, target: SocketAddr, socket: TcpSocket) -> io::Result<()> {
        self.node.connect_using_socket(target, socket).await
    }
```

src/tests/performance/connections.rs
```diff
@@ -170,15 +170,34 @@ async fn p002_connections_load() {
...
+            socket.set_reuseport(true).unwrap();
+
+            // If there is address for our thread in the pool we can use it.
+            // Otherwise we'll not set bound_addr and use local IP addr (127.0.0.1).
+            if IPS.len() > i as usize {
+                let source_addr = SocketAddr::new(
+                    IpAddr::V4(Ipv4Addr::from_str(IPS[i as usize]).unwrap()),
+                    CONNECTION_PORT + port_idx,
+                );
+                port_idx += 1;
```
Rqnsom, line 191: Shouldn't we do this in the outer for loop (which starts with line 175) 🤔

Rqnsom: Or why do we need to change port_idx at all?
Didn't we say that restarting the rippled node after every subtest helps with the reuseaddr issue?

ljedrz (commented):
Update-wise, I'd expect to see calls to `Node::start_listening` in all the tests where the synthetic node is the one being connected to; see [the commit that added it](https://github.com/ljedrz/pea2pea/commit/cb684cc6e8617e13c1dd16c2f17ba5dce1585743) to see how many changes in tests it required upstream.

src/tests/performance/connections.rs
```diff
@@ -199,15 +218,15 @@ async fn p002_connections_load() {
             let (exit_tx, exit_rx) = tokio::sync::oneshot::channel::<()>();
             synth_exits.push(exit_tx);
 
+            let sock = synth_sockets.remove(0);
```
Rqnsom, line 221: I just figured,
you could use this as an alternative approach to this for loop:

```rust
        // start synthetic nodes
        for socket in synth_sockets.into_iter() {
            let (exit_tx, exit_rx) = tokio::sync::oneshot::channel::<()>();
            synth_exits.push(exit_tx);

            let synth_handshaken = handshake_tx.clone();
            // Synthetic node runs until it completes or is instructed to exit
            synth_handles.push(tokio::spawn(async move {
                tokio::select! {
                    _ = exit_rx => {},
                    _ = simulate_peer(node_addr, synth_handshaken, socket) => {},
                };
            }));
```
to avoid having that unused `underscore` in `for _ in 0..synth_count` 😎
and avoid doing manually this part: `synth_sockets.remove(0);`

Rqnsom (commented):
> Update-wise, I'd expect to see calls to Node::start_listening in all the tests where the synthetic node is the one being connected to; see [the commit that added it](https://github.com/ljedrz/pea2pea/commit/cb684cc6e8617e13c1dd16c2f17ba5dce1585743) to see how many changes in tests it required upstream.

Yeah, good point, I know some tests were failing in the Algorand repo before those changes were applied. Are tests passing here without that change in place?

In any case, that could be done in a separate PR which could be delivered first, would be easier to review

PR 1 -> a simple uplift to 0.43 and make sure that everything works as expected
PR 2 -> start using new features from 0.43 (basically this PR)

asmie:
Splitting this PR into PR1 and PR2 would result in PR1 to be unbuildable/not working correctly as eg. perf tests need to have features from PR2.
I'll add review notes to this PR.

src/tools/inner_node.rs
asmie: I'll change it. I've added it the same way that was connect() implemented (to be consistent with other code in inner_node.rs).

src/tests/performance/connections.rs
asmie: Changed to restart node each subset iteration. port_idx is eliminated.

src/tests/performance/connections.rs
asmie: True, changed this way :)

src/tools/inner_node.rs
Rqnsom: Aha, okay. Feel free to apply this improvement to that function as well (in some future PR/commit)

Rqnsom (commented):
LGTM for the current part,
but please double-check that all other tests are passing (like the handshake test C002)

See the algorand adaptation [here](https://github.com/runziggurat/algorand/blob/main/src/tests/conformance/handshake.rs#L57)

src/tests/conformance/cmd/cluster.rs
```diff
@@ -26,15 +26,13 @@ async fn c024_TM_CLUSTER_node_should_connect_to_other_nodes_in_cluster() {
     test_config.pea2pea_config.desired_listening_port = Some(DEFAULT_PORT);
     test_config.synth_node_config.generate_new_keys = false;
     let mut synth_node = SyntheticNode::new(&test_config).await;
+    let listening_addr = synth_node.start_listening().await.unwrap();
```
lwojtow, line 29: `.expect` here?

src/tests/performance/connections.rs
```diff
@@ -161,24 +161,40 @@ async fn p002_connections_load() {
...
+        let target = TempDir::new().expect("couldn't create a temporary directory");
+        // start node
+        let mut node = Node::builder()
+            .max_peers(MAX_PEERS as usize)
+            .start(target.path(), NodeType::Stateless)
+            .await
+            .unwrap();
+        let node_addr = node.addr();
+
+        let mut synth_sockets = Vec::with_capacity(synth_count.into());
```
lwojtow, line 174: `into` here looks silly as it should be usize from the start.

src/tests/performance/connections.rs
```diff
@@ -187,15 +203,15 @@ async fn p002_connections_load() {
...
+        let mut synth_handles = Vec::with_capacity(synth_count.into());
+        let mut synth_exits = Vec::with_capacity(synth_count.into());
         let (handshake_tx, mut handshake_rx) =
             tokio::sync::mpsc::channel::<()>(synth_count as usize);
 
         let test_start = Instant::now();
 
         // start synthetic nodes
-        for _ in 0..synth_count {
+        for socket in synth_sockets.into_iter() {
```
lwojtow, line 214: Why is `into_iter` needed?

ljedrz (approved):
LGTM :ok_hand:

src/tests/performance/connections.rs
Rqnsom: Ah, right, `for` loops use `into_iter` by default

src/tests/conformance/cmd/cluster.rs
asmie: Changed.

src/tests/performance/connections.rs
```diff
@@ -157,28 +157,45 @@ async fn p002_connections_load() {
...
+            // If there is address for our thread in the pool we can use it.
+            // Otherwise we'll not set bound_addr and use local IP addr (127.0.0.1).
+            let ip = if let Some(ip_addr) = ips.pop() {
+                SocketAddr::new(
+                    IpAddr::V4(Ipv4Addr::from_str(ip_addr).unwrap()),
+                    CONNECTION_PORT,
+                )
+            } else {
+                "127.0.0.1:0".parse().unwrap()
+            };
```
Rqnsom, line 193: noice

Rqnsom (approved):
LGTM 🚀

PR #95, opened 2022-12-12
feat: getting transaction details performance test

asmie:
Performance test that is done by getting transaction details (using transaction hash) from the node.

### DIFF
--- a/src/tests/performance/get_trans.rs
+++ b/src/tests/performance/get_trans.rs
@@ -180,13 +180,11 @@ async fn simulate_peer(node_addr: SocketAddr, socket: TcpSocket, tx_hash: [u8; T
         .await
         .expect("unable to connect to node");
 
-    let mut seq: u32 = 1;
-
-    for _ in 0..REQUESTS {
+    for seq in 0..REQUESTS {
         let payload = Payload::TmGetObjectByHash(TmGetObjectByHash {
             r#type: ObjectType::OtTransactions as i32,
             query: true,
-            seq: Some(seq),
+            seq: Some(seq as u32),
             ledger_hash: None,
             fat: None,
             objects: vec![TmIndexedObject {
@@ -197,8 +195,6 @@ async fn simulate_peer(node_addr: SocketAddr, socket: TcpSocket, tx_hash: [u8; T
                 ledger_seq: None,
             }],
         });
-
-        seq += 1;
 
         // Query transaction via peer protocol.
         if synth_node.is_connected(node_addr) {
--- a/src/tests/performance/get_trans.rs
+++ b/src/tests/performance/get_trans.rs
@@ -1,0 +1,243 @@
+use std::{
+    io::ErrorKind,
+    net::{IpAddr, Ipv4Addr, SocketAddr},
+    str::FromStr,
+    time::{Duration, Instant},
+};
+
+use tempfile::TempDir;
+use tokio::net::TcpSocket;
+
+use crate::{
+    protocol::{
+        codecs::message::Payload,
+        proto::{
+            tm_get_object_by_hash::ObjectType, TmGetObjectByHash, TmIndexedObject, TmTransactions,
+        },
+    },
+    setup::node::{Node, NodeType},
+    tools::{
+        config::TestConfig,
+        constants::{EXPECTED_RESULT_TIMEOUT, TEST_ACCOUNT},
+        ips::IPS,
+        metrics::{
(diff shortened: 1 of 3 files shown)


src/tests/performance/get_trans.rs
```diff
@@ -0,0 +1,243 @@
...
+    }
+
+    // Display results table
+    println!("\r\n{}", table);
+}
+
+async fn simulate_peer(node_addr: SocketAddr, socket: TcpSocket, tx_hash: [u8; 32]) {
+    let config = TestConfig::default();
+
+    let mut synth_node = SyntheticNode::new(&config).await;
```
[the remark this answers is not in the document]
asmie: Done.

src/tests/performance/get_trans.rs
```diff
@@ -0,0 +1,241 @@
...
+        let target = TempDir::new().expect("Unable to create TempDir");
+        let mut node = Node::builder()
+            .max_peers(MAX_PEERS)
+            .start(target.path(), NodeType::Stateful)
+            .await
+            .unwrap();
+        let node_addr = node.addr();
+
+        let mut synth_sockets = Vec::with_capacity(synth_count);
+        let mut ips = IPS.to_vec();
```
ljedrz, line 94: nit: since we only `pop` from the list of `ips`, we might as well iterate as follows:
```
for (_, ip_addr) in (0..synth_count).zip(IPS)
```

src/tests/performance/get_trans.rs
```diff
@@ -0,0 +1,241 @@
...
+
+    // Establish peer connection
+    synth_node
+        .connect_from(node_addr, socket)
+        .await
+        .expect("unable to connect to node");
+
+    let mut seq: u32 = 1;
+
+    for _ in 0..REQUESTS {
```
ljedrz, line 185: why not `for seq in 1..=REQUESTS`?

ljedrz (approved):
LGTM

src/tests/performance/get_trans.rs
asmie: The IPS list can be empty or contain less entries than we'd like to spawn threads. In such case we use localhost addresses. When iterators are zipped the loop would not work if IPS is empty and not produce localhost addresses.

src/tests/performance/get_trans.rs
asmie: Changed.

ljedrz: ah, so it can start from `0` as well? cool, even simpler

asmie: Yes, it can, as in rippled code (PeerImpl.cpp:2680) it's just simple:
```
if (packet.has_seq())
     reply.set_seq(packet.seq());
```
so we can use zero. Initially used one as it was in c007_TM_GET_OBJECT_BY_HASH_get_transaction_by_hash packet construction code. And as there is only one packet constructed it does not matter if there is one, zero or 666 :)

src/tests/performance/get_trans.rs
```diff
@@ -0,0 +1,237 @@
...
+
+        // Query transaction via peer protocol.
+        if synth_node.is_connected(node_addr) {
+            synth_node
+                .unicast(node_addr, payload)
+                .expect("unable to send message");
+        } else {
+            synth_node.shut_down().await;
+            return;
+        }
```
Rqnsom, line 207: `else` block could be removed (and a simple break would lead to a shutdown and function return):
```rust
        // Query transaction via peer protocol.
        if !synth_node.is_connected(node_addr) {
            break;
        }

        synth_node
            .unicast(node_addr, payload)
            .expect("unable to send message");
```

src/tests/performance/get_trans.rs
```diff
@@ -0,0 +1,237 @@
...
+                    }
+                }
+                Err(e) => match e.kind() {
+                    ErrorKind::TimedOut => {
+                        break;
+                    }
+                    _ => {
+                        panic!("Unexpected error: {:?}", e);
+                    }
+                },
```
Rqnsom, line 231: could be shorter:
```rust
                Err(e) if e.kind() == ErrorKind::TimedOut => break,
                Err(e) => panic!("Unexpected error: {:?}", e),
```

Rqnsom (commented):
late check now that I have time... sorry, I had a busy day 😀

I just left a two possible future improvements, it's LGTM

PR #99, opened 2022-12-22
feat: add basic network crawling

lwojtow:
Initial idea for network crawling.

### DIFF
--- a/src/tools/crawler/crawler.rs
+++ b/src/tools/crawler/crawler.rs
@@ -1,0 +1,106 @@
+use std::{net::SocketAddr, str::FromStr, sync::Arc};
+
+use pea2pea::protocols::{Handshake, Reading, Writing};
+use tokio::{
+    sync::mpsc::{channel, Receiver},
+    time::{timeout, Instant},
+};
+use tracing::{error, info, trace, warn};
+use ziggurat_xrpl::{
+    protocol::{
+        codecs::message::{BinaryMessage, Payload},
+        proto::TmEndpoints,
+    },
+    setup::constants::CONNECTION_TIMEOUT,
+    tools::inner_node::InnerNode,
+};
+
+use crate::network::KnownNetwork;
+
+pub(super) struct Crawler {
+    node: InnerNode,
+    known_network: Arc<KnownNetwork>,
+    receiver: Receiver<(SocketAddr, BinaryMessage)>,
+}
+
+impl Crawler {
+    pub(super) async fn new() -> Self {
+        let (sender, receiver) = channel(1024);
+        let node = InnerNode::new(&Default::default(), sender).await;
+        node.enable_handshake().await;
+        node.enable_reading().await;
+        node.enable_writing().await;
+        let known_network = Arc::new(Default::default());
+        Self {
+            node,
+            known_network,
+            receiver,
+        }
+    }
+
+    pub(super) async fn start_processing(&mut self) {
+        while let Some((addr, message)) = self.receiver.recv().await {
+            if let Payload::TmEndpoints(endpoints) = message.payload {
+                self.process_endpoints_message(addr, endpoints).await;
+            }
+        }
+    }
+
+    async fn process_endpoints_message(&mut self, addr: SocketAddr, endpoints: TmEndpoints) {
+        info!(
+            "Received endpoints from {}: {:?}",
+            addr, endpoints.endpoints_v2
+        );
+        for endpoint in endpoints.endpoints_v2 {
+            match SocketAddr::from_str(&endpoint.endpoint) {
+                Ok(peer) => {
+                    self.process_peer(addr, endpoint.hops, peer).await;
+                }
+                Err(e) => {
+                    error!("invalid address from {}: {}", addr, e);
+                }
+            }
+        }
+    }
+
+    async fn process_peer(&mut self, addr: SocketAddr, hops: u32, peer: SocketAddr) {
+        self.known_network.insert_node(peer).await;
+        if hops == 0 {
+            self.known_network.insert_connection(addr, peer).await;
+        }
+        self.get_peers(peer).await;
+    }
+
+    pub(super) async fn get_peers(&self, addr: SocketAddr) {
+        let node = self.node.clone();
+        let network = self.known_network.clone();
+
+        tokio::spawn(async move {
+            network.insert_node(addr).await;
+            let start = Instant::now();
+            if connect_node(node, addr).await {
+                network.set_connected(addr, start.elapsed()).await;
+            }
+        });
+    }
+}
+
+async fn connect_node(node: InnerNode, addr: SocketAddr) -> bool {
+    let connection_result = timeout(CONNECTION_TIMEOUT, node.connect(addr)).await;
+    match connection_result {
+        Ok(response) => match response {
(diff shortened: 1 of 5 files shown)


src/tools/crawler/main.rs
```diff
@@ -25,7 +27,12 @@ fn start_logger(default_level: LevelFilter) {
 
 #[tokio::main]
 async fn main() {
-    start_logger(LevelFilter::DEBUG);
+    start_logger(LevelFilter::INFO);
     let args = Args::parse();
-    debug!("Crawler starting with args: {:?}", args);
+    info!("Crawler starting with args: {:?}", args);
+    let mut crawler = Crawler::new().await;
```
Rqnsom, line 33: could we add some newlines, it's so tight without any

src/tools/crawler/crawler.rs
```diff
@@ -0,0 +1,107 @@
...
+                    self.process_peer(addr, endpoint.hops, peer).await;
+                }
+                Err(e) => {
+                    warn!("invalid address from {}: {}", addr, e);
+                }
+            }
+        }
+    }
+
+    async fn process_peer(&mut self, addr: SocketAddr, hops: u32, peer: SocketAddr) {
```
Rqnsom, line 66:
```suggestion
    async fn process_peer(&mut self, addr: SocketAddr, hops: u32, addr: SocketAddr) {
```

src/tools/crawler/crawler.rs
```diff
@@ -0,0 +1,107 @@
...
+        }
+    }
+
+    async fn process_endpoints_message(&mut self, addr: SocketAddr, endpoints: TmEndpoints) {
+        info!(
+            "Received endpoints from {}: {:?}",
+            addr, endpoints.endpoints_v2
+        );
+        for endpoint in endpoints.endpoints_v2 {
+            match SocketAddr::from_str(&endpoint.endpoint) {
```
Rqnsom, line 55: at this point (after CPR tests are done), I think we can use expect here
if the node sends an invalid address, they have a bug and the crawler should just panic

src/tools/crawler/network.rs
```diff
@@ -0,0 +1,101 @@
...
+    connections: RwLock<HashSet<KnownConnection>>,
+}
+
+impl KnownNetwork {
+    pub(super) async fn insert_node(&self, addr: SocketAddr) {
+        let mut nodes = self.nodes.write().await;
+        if !nodes.contains_key(&addr) {
+            nodes.insert(addr, KnownNode::default());
+            info!("Known nodes: {}", nodes.len());
+        }
```
Rqnsom, line 24: we might just use the [insert](https://doc.rust-lang.org/std/collections/struct.HashMap.html#method.insert) fn directly and use the return value to printout the info trace

src/tools/crawler/network.rs
```diff
@@ -0,0 +1,101 @@
...
+
+impl PartialEq for KnownConnection {
+    fn eq(&self, other: &Self) -> bool {
+        let (a, b) = (self.a, self.b);
+        let (c, d) = (other.a, other.b);
+
+        a == d && b == c || a == c && b == d
+    }
+}
+/// A node encountered in the network or obtained from one of the peers.
```
Rqnsom, line 89:
```suggestion
}

/// A node encountered in the network or obtained from one of the peers.
```

src/tools/crawler/network.rs
```diff
@@ -0,0 +1,101 @@
...
+        match a.cmp(&b) {
+            Ordering::Greater => {
+                b.hash(state);
+                a.hash(state);
+            }
+            _ => {
+                a.hash(state);
+                b.hash(state);
+            }
+        }
```
Rqnsom, line 67: maybe 
```suggestion
        if a.cmp(&b) == Ordering::Greater {
            b.hash(state);
            a.hash(state);
        } else {
            a.hash(state);
            b.hash(state);
        }
```

Rqnsom (approved):
LGTM overall, but might take another look tomorrow

asmie (approved):

src/tools/crawler/crawler.rs
```diff
@@ -0,0 +1,107 @@
...
+    pub(super) async fn get_peers(&self, addr: SocketAddr) {
+        let node = self.node.clone();
+        let network = self.known_network.clone();
+
+        tokio::spawn(async move {
+            network.insert_node(addr).await;
+            let start = Instant::now();
+            if connect_node(node, addr).await {
+                let handshake_time = start.elapsed();
+                network.set_connected(addr, handshake_time).await;
```
kylegranger, line 83: maybe
```
network.set_connected(addr, start.elapsed()).await;
```

kylegranger (approved):

src/tools/crawler/network.rs
lwojtow: True, but the thing is I don't want to override the struct in the map as it contains some stats.

Rqnsom: right! then try_insert would be actually useful here, but a stable feature yet

src/tools/crawler/main.rs
lwojtow: Done

src/tools/crawler/crawler.rs
lwojtow: To be honest I prefer this name. With this change we would have two `addr` params.

src/tools/crawler/crawler.rs
lwojtow: I would suggest to leave it as is for the time being - somehow I feel that panicking (and killing whole crawler) just because someone sent us data we can parse would be inappropriate here.

src/tools/crawler/network.rs
lwojtow: Done

src/tools/crawler/network.rs
lwojtow: That's the way it's done in zcash, so let's keep it consistent.

src/tools/crawler/crawler.rs
lwojtow: Done

src/tools/crawler/crawler.rs
Rqnsom: At least let's change the error trace to `error` - since it's a potentially big issue on their side, no?

The whole crawler work is not really important if their nodes sometimes send invalid addresses randomly...

src/tools/crawler/crawler.rs
Rqnsom: Sure!
I know that I was confused by the `peer` name when I saw it, so I had to check what it meant - basically, it took me more time to understand the code, and this suggestion would probably help me to understand it faster, but it what works best for me doesn't necessarily work for others 😅

Rqnsom (approved):

src/tools/crawler/crawler.rs
lwojtow: Let's leave it as peer for now and see if anything better comes up in the future.

src/tools/crawler/crawler.rs
lwojtow: Done. Although I think we have to formalize distinction between error/warn levels for the future.

PR #100, opened 2022-12-23
feat: crawler connected stats

lwojtow:
* Small improvement: keep track of which nodes are already connected and don't connect again
* Add connection time to node's stats.

### DIFF
--- a/src/tools/crawler/network.rs
+++ b/src/tools/crawler/network.rs
@@ -7,7 +7,7 @@ use std::{
 };
 
 use tokio::{sync::RwLock, time::Instant};
-use tracing::{error, info};
+use tracing::info;
 
 #[derive(Default)]
 pub struct KnownNetwork {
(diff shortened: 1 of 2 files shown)


src/tools/crawler/network.rs
```diff
@@ -34,7 +39,9 @@ impl KnownNetwork {
         if !nodes.contains_key(&addr) {
             error!("set_connected(): addr not found ({})", addr);
         } else {
-            nodes.get_mut(&addr).unwrap().handshake_time = Some(handshake_time);
+            let mut node = nodes.get_mut(&addr).unwrap();
```
ljedrz, line 42: why aren't we doing `if let Some(node) = nodes.get_mut(&addr)` as the first branch? we would avoid a double query that way

src/tools/crawler/network.rs
```diff
@@ -16,11 +16,16 @@ pub struct KnownNetwork {
 }
 
 impl KnownNetwork {
-    pub(super) async fn insert_node(&self, addr: SocketAddr) {
+    /// Inserts addr to known_nodes if not yet present (so to avoid overriding the node's statistics).
+    /// Returns true if it's a new node, false otherwise.
+    pub(super) async fn insert_node(&self, addr: SocketAddr) -> bool {
```
ljedrz, line 21: for later: haven't dug deeper, but perhaps we could use the `Entry` API (with `or_default`) to avoid a double query here

ljedrz (approved):
haven't dug too deep, but these changes look good to me; left some suggestions not strictly related to this PR

src/tools/crawler/network.rs
lwojtow: Done

src/tools/crawler/network.rs
lwojtow: I will check this later, thanks for the idea.

PR #101, opened 2023-01-05
feat: use /crawl endpoint for crawling ripple network

lwojtow:
Crawler using /crawl endpoint to find other peers. More information at https://xrpl.org/peer-crawler.html

### DIFF
--- a/src/tools/crawler/crawl.rs
+++ b/src/tools/crawler/crawl.rs
@@ -1,0 +1,136 @@
+//! Contains structs and methods to crawl ripple network according to instruction at https://xrpl.org/peer-crawler.html
+
+use std::{
+    net::{IpAddr, SocketAddr},
+    num::ParseIntError,
+    time::Duration,
+};
+
+use reqwest::{Client, StatusCode};
+use serde::Deserialize;
+use thiserror::Error;
+use tokio::time::Instant;
+
+const CRAWLER_DEFAULT_PORT: u16 = 51235;
+
+#[allow(dead_code)]
+#[derive(Debug, Deserialize)]
+pub(super) struct Peer {
+    pub(super) complete_ledgers: Option<String>,
+    pub(super) complete_shards: Option<String>,
+    pub(super) ip: Option<String>,
+    pub(super) port: Option<Port>,
+    pub(super) public_key: String,
+    #[serde(rename = "type")]
+    pub(super) connection_type: String,
+    pub(super) uptime: u32,
+    pub(super) version: String,
+}
+
+impl Peer {
+    /// Returns port number for the peer. If no port was present in the response then returns the default port.
+    /// Returns [ParseIntError] if the response contained invalid value.
+    pub(super) fn port(&self) -> Result<u16, ParseIntError> {
+        self.port
+            .as_ref()
+            .map(|p| match p {
+                Port::Number(n) => Ok(*n),
+                Port::String(s) => s.parse(),
+            })
+            .unwrap_or(Ok(CRAWLER_DEFAULT_PORT))
+    }
+}
+
+#[derive(Deserialize)]
+pub(super) struct CrawlResponse {
+    pub(super) overlay: Overlay,
+    pub(super) server: Server,
+}
+
+#[derive(Deserialize)]
+pub(super) struct Overlay {
+    pub(super) active: Vec<Peer>,
+}
+
+#[allow(dead_code)]
+#[derive(Deserialize)]
+pub(super) struct Server {
+    pub(super) build_version: String,
+    pub(super) server_state: String,
+    pub(super) uptime: u32,
+}
+
+#[derive(Debug, Deserialize)]
+#[serde(untagged)]
+pub(super) enum Port {
+    Number(u16),
+    String(String),
+}
+
+/// Connects to `https://IP:PORT/crawl` to query `addr's` peers.
+/// On success returns the response and the time it took to connect, send the request and read the response.
+/// On failure it returns a [CrawlError].
+pub(super) async fn get_crawl_response(
+    client: Client,
+    addr: SocketAddr,
+) -> Result<(CrawlResponse, Duration), CrawlError> {
+    let url = format!("https://{}:{}/crawl", format_ip_for_url(addr), addr.port());
+    let start = Instant::now();
+    let response = client
+        .get(url)
+        .send()
+        .await
+        .map_err(|e| CrawlError::Connection(e.to_string()))?;
+    let elapsed = start.elapsed();
+    if response.status() == StatusCode::OK {
+        let response = response
+            .json::<CrawlResponse>()
+            .await
+            .map_err(|e| CrawlError::Response(e.to_string()))?;
+        Ok((response, elapsed))
+    } else {
+        Err(CrawlError::Response(format!(
+            "status: {}",
+            response.status()
+        )))
+    }
+}
+
+/// Formats ip address to be used in http url format.
+/// That means that IPv6 address is wrapped in []
+fn format_ip_for_url(addr: SocketAddr) -> String {
+    if let IpAddr::V6(ip) = addr.ip() {
+        format!("[{}]", ip)
+    } else {
+        addr.ip().to_string()
+    }
+}
+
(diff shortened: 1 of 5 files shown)


src/tools/crawler/crawl.rs
```diff
@@ -0,0 +1,121 @@
...
+
+use reqwest::{Client, StatusCode};
+use serde::Deserialize;
+use thiserror::Error;
+use tokio::time::Instant;
+
+const CRAWLER_DEFAULT_PORT: u16 = 51235;
+const CRAWLER_TIMEOUT: Duration = Duration::from_secs(10);
+
+#[allow(dead_code)]
```
ljedrz, line 17: is the whole crawler module not feature-gated?

src/tools/crawler/crawl.rs
```diff
@@ -0,0 +1,121 @@
...
+    /// Returns port number for the peer. If no port was present in the response then returns the default port.
+    /// Returns [ParseIntError] if the response contained invalid value.
+    pub(super) fn port(&self) -> Result<u16, ParseIntError> {
+        self.port
+            .as_ref()
+            .map(|p| match p {
+                Port::Number(n) => Ok(*n),
+                Port::String(s) => s.parse(),
+            })
+            .unwrap_or(Ok(CRAWLER_DEFAULT_PORT))
```
ljedrz, line 41: this means an error will actually never be returned, so the return value either should not be a `Result` or there should be a `?` after the call to `parse`

src/tools/crawler/crawl.rs
```diff
@@ -0,0 +1,121 @@
...
+}
+
+/// Connects to `https://IP:PORT/crawl` to query `addr's` peers.
+/// Returns (response, time it took to connect) on success, or
+/// [CrawlError] when unable to connect.
+pub(super) async fn get_crawl_response(
+    addr: SocketAddr,
+) -> Result<(CrawlResponse, Duration), CrawlError> {
+    let url = format!("https://{}:{}/crawl", format_ip_for_url(addr), addr.port());
+    let client = Client::builder()
```
ljedrz, line 78: how often is this function expected to be called? it might be preferable for the `Client` to persist if it happens quite often

src/tools/crawler/crawl.rs
```diff
@@ -0,0 +1,121 @@
...
+        .danger_accept_invalid_certs(true)
+        .timeout(CRAWLER_TIMEOUT)
+        .build()
+        .expect("unable to build web client");
```
ljedrz, line 82:
```suggestion
        .expect("unable to build the web client");
```

src/tools/crawler/crawler.rs
```diff
@@ -1,107 +1,100 @@
...
-            receiver,
-        }
-    }
-
-    pub(super) async fn start_processing(&mut self) {
-        while let Some((addr, message)) = self.receiver.recv().await {
-            if let Payload::TmEndpoints(endpoints) = message.payload {
-                self.process_endpoints_message(addr, endpoints).await;
-            }
+            known_network: Arc::new(Default::default()),
```
ljedrz, line 23:
```suggestion
            known_network: Default::default(),
```

src/tools/crawler/crawler.rs
```diff
@@ -1,107 +1,100 @@
...
-                    network.set_connected(addr, start.elapsed()).await;
-                }
+/// Processes /crawl response: updates `known_network` and starts more crawling recursively.
+fn process_response(
+    from: SocketAddr,
+    response: CrawlResponse,
+    connecting_time: Duration,
+    known_network: Arc<KnownNetwork>,
+) -> BoxFuture<'static, ()> {
+    // Wrapped in BoxFuture to allow for async recursion.
```
ljedrz, line 54: where does the recursion occur?

src/tools/crawler/crawler.rs
```diff
@@ -1,107 +1,100 @@
...
+    async move {
+        known_network
+            .update_stats(from, connecting_time, response.server.build_version)
+            .await;
+        for peer in response.overlay.active {
+            if let Some(addr) = parse_peer_addr(from, peer) {
+                known_network.insert_connection(from, addr).await;
```
ljedrz, line 61: wouldn't it make sense to insert all the new connections based on the response first, and only then to crawl?

src/tools/crawler/crawler.rs
```diff
@@ -1,107 +1,100 @@
...
-                trace!("connected to {}", addr);
-                true
-            }
-            Err(e) => {
-                warn!("unable to connect to {} due to {}", addr, e);
-                false
+/// Tries to parse address information from response.
+/// On success returns Some(SocketAddr) on failure logs error and returns None.
+fn parse_peer_addr(from: SocketAddr, peer: Peer) -> Option<SocketAddr> {
+    let ip = match peer.ip.as_ref() {
```
ljedrz, line 72: do we desire such level of error detail? I'd just go with https://doc.rust-lang.org/std/net/enum.SocketAddr.html#impl-FromStr-for-SocketAddr for the entire address at once

ljedrz (commented):
There's plenty going on in this PR; I'd prefer to see it split into several commits.

Left some comments/suggestions too.

src/tools/crawler/crawl.rs
lwojtow: It is. This `allow(dead_code)` is about (yet) unused fields in `Peer`

src/tools/crawler/crawl.rs
lwojtow: I've added a test to show when error is returned (bottom of the file).

ljedrz: ah, I missed the `as_ref`; ok, so it's an `Option<T>` turned into an `Option<Result<T>>`, unwrapped from the initial option at the end

lwojtow: Yes. Port can omitted in the response in which case I'm using the default. If there is a string - it tries to parse it.

src/tools/crawler/crawl.rs
lwojtow: Done

src/tools/crawler/crawler.rs
lwojtow: process_response -> crawl -> process_response

src/tools/crawler/crawler.rs
ljedrz: I was thinking that we could insert connections wholesale, reducing lock contention, and only start crawling afterwards

src/tools/crawler/crawl.rs
lwojtow: Done. It required some more changes but it's better now, thanks.

src/tools/crawler/crawler.rs
lwojtow: I could do:
`SocketAddr::from_str(format!("{}:{}", peer.ip.as_ref()?, peer.port().ok()?).as_str()).ok()`
and be done, but then we lose information why it didn't work, so we can't log it.
Is that ok?

src/tools/crawler/crawl.rs
```diff
@@ -0,0 +1,141 @@
...
+
+#[derive(Debug, Deserialize)]
+#[serde(untagged)]
+pub(super) enum Port {
+    Number(u16),
+    String(String),
+}
+
+/// Connects to `https://IP:PORT/crawl` to query `addr's` peers.
+/// Returns (response, time it took to connect) on success, or
```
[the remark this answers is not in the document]
lwojtow: Yes, will adjust the comment explaining this.

asmie (approved):

src/tools/crawler/crawler.rs
ljedrz: hence my question about the level of detail we need; for a crawler it is likely that we'll typically have dozens - if not hundreds - of connections at a time, so any individual connection errors are likely to be of little interest; that being said I have no strong feelings here

lwojtow: Agreed and done.

kylegranger (approved):

PR #102, opened 2023-01-09
feat: add connection retry if failed

lwojtow:
Try to connect in a loop. If failed to connect wait for 10 seconds and try again. Maximum number of tries set to 255.

### DIFF
--- a/src/tools/crawler/network.rs
+++ b/src/tools/crawler/network.rs
@@ -50,6 +50,14 @@ impl KnownNetwork {
         node.connecting_time = Some(connecting_time);
         node.server = Some(server_version);
     }
+
+    /// Increases connection failures to the `addr` and returns its new value.
+    pub(super) async fn increase_connection_failures(&self, addr: SocketAddr) -> u8 {
+        let mut nodes = self.nodes.write().await;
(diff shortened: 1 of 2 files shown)


src/tools/crawler/network.rs
```diff
@@ -50,6 +50,16 @@ impl KnownNetwork {
         node.connecting_time = Some(connecting_time);
         node.server = Some(server_version);
     }
+
+    /// Increases connection failures to the `addr` and returns its new value.
+    pub(super) async fn increase_connection_failures(&self, addr: SocketAddr) -> u8 {
+        let mut nodes = self.nodes.write().await;
+        let mut node = nodes.get_mut(&addr).unwrap();
+        if node.connection_failures < u8::MAX {
```
asmie, line 58: Instead of whole `if` block we could use `saturating_add` of u8 type (https://doc.rust-lang.org/std/primitive.u8.html#method.saturating_add).

asmie (approved):
LGTM with one comment.

src/tools/crawler/network.rs
lwojtow: Done.

PR #106, opened 2023-01-12
refactor: use ziggurat-core-* packages

asmie:
This is a cumulative commit for various elements done during update xrpl to use ziggurat core packages.
Changes done:
- updated ips.py script to be in the newest version;
- removed code that is already present in ziggurat-core-metrics;
- removed code that is already present in ziggurat-core-utils;
- made performance tests to be run as a feature;
- reformatted source files to have consistent \r\n formatting.


tools/ips.py
```diff
@@ -4,30 +4,32 @@
 # writes that list to the Rust file.
 # When testing is done, just change --op to 'remove' and run the script again to remove all the dummy devices (or aliases).
 #
-# Linux and Mac supported! Script need to be run with sudo.
+# Linux and Mac supported! User must be in sudoers file to use this script.
```
Rqnsom, line 7: Should we move this script to the ziggurat-core repo?

Rqnsom (approved):
LGTM,

it would be nice to break these into smaller PRs
now we have here lots of different stuff in here (and no PR description):
- script update
- applying ziggurat-core-metrics
- applying ziggurat-core-utils
- adding performance feature

tools/ips.py
asmie: Well, we could but it would demand user to fetch core repo, extract ips.py file, copy it to folder with repository and then run it. Till we think out the easier way I'd leave it here.

PR #107, opened 2023-01-12
feat: add option to dump crawler's stats to a file

lwojtow:
Adds option to dump crawler's stats to a file.
At this point there are no safety guards to protect against directory traversal.

### DIFF
--- a/src/tools/crawler/rpc.rs
+++ b/src/tools/crawler/rpc.rs
@@ -1,5 +1,8 @@
 use std::{
+    fs,
     net::SocketAddr,
+    ops::Deref,
+    path::PathBuf,
     sync::{Arc, Mutex},
 };
 
@@ -7,7 +10,8 @@ use jsonrpsee::{
     server::{ServerBuilder, ServerHandle},
     RpcModule,
 };
-use tracing::debug;
+use serde::Deserialize;
+use tracing::{debug, warn};
 
 use crate::metrics::NetworkSummary;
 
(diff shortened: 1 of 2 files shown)


src/tools/crawler/rpc.rs
```diff
@@ -34,9 +38,24 @@ pub async fn initialize_rpc_server(rpc_addr: SocketAddr, rpc_context: RpcContext
 fn create_rpc_module(rpc_context: RpcContext) -> RpcModule<RpcContext> {
     let mut module = RpcModule::new(rpc_context);
     module
-        .register_method("getmetrics", |_, rpc_context| {
+        .register_method("getmetrics", |params, rpc_context| {
+            let report_params = params.parse::<ReportParams>()?;
+            if let Some(path) = report_params.file {
+                let content = serde_json::to_string(rpc_context.0.lock().unwrap().deref())?;
+                // TODO: consider some checks against directory traversal
```
asmie, line 45: To check if someone is not trying to put parent directory somewhere we can just use:
```
if pathBuf.components().into_iter().any(|p| p == Component::ParentDir) {
   // directory traversal here...
}
```
Or check if there are any other '/' symbols (especially on the beginning).

asmie (approved):
LGTM

src/tools/crawler/rpc.rs
lwojtow: Yes, disallowing any relative path is one way. Also, we could do `canonicalize` and check if the final path is somewhere we allow to be. Or, take only filename part and dump the file either in /tmp or current dir.
Let's wait and see how this file is going to be read, then we'll figure out how to do this safely.

PR #108, opened 2023-01-13
feat: make crawler to update its stats in a loop

lwojtow:
* Increase retry delay after failure
* Make crawler update stats in a loop
* Reset connection failure count when successfully connected


Rqnsom (approved):
LGTM

PR #110, opened 2023-01-17
Feat/crawler show node ids

lwojtow:
* extract good nodes ids
* fix clippy warning


kylegranger (approved):
👍  nice, esp. having the node filtering and setting the node_ids in separate functions.  This is something I should do in the zcash crawler (next round)

Rqnsom (approved):
LGTM

PR #112, opened 2023-01-20
chore(crawler): randomize time periods for requests

Rqnsom:
- uniformly spread the peerlist requests in the time domain
- also increase request frequency

### DIFF
--- a/src/tools/crawler/crawler.rs
+++ b/src/tools/crawler/crawler.rs
@@ -2,6 +2,7 @@ use std::{net::SocketAddr, str::FromStr, sync::Arc, time::Duration};
 
 use futures_util::{future::BoxFuture, FutureExt};
 use pea2pea::protocols::Handshake;
+use rand::Rng;
 use reqwest::Client;
 use tokio::time::sleep;
 use tracing::{trace, warn};
@@ -12,7 +13,8 @@ use crate::{
     network::KnownNetwork,
 };
 
-const DELAY_AFTER_CONNECTION: Duration = Duration::from_secs(60 * 5);
+const CONNECTION_RETRY_MIN_SEC: u64 = 3 * 60; // 3 minutes
+const CONNECTION_RETRY_MAX_SEC: u64 = 5 * 60; // 5 minutes
 
 pub(super) struct Crawler {
     pub(super) known_network: Arc<KnownNetwork>,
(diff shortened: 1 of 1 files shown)


Rqnsom (commented):
`let duration = rand::thread_rng().gen_range(5..=DELAY_AFTER_CONNECTION);` wher DELAY_AFTER_CONNECTION=60

We can discuss these constants, but these values worked well for me

src/tools/crawler/crawler.rs
```diff
@@ -36,21 +37,27 @@ pub(super) fn crawl(
...
+                    if failures == u8::MAX {
+                        warn!("Giving up connecting to {addr}");
+                        break;
                     }
-                    // Even if connection was successful try again after a while to update peers.
-                    sleep(DELAY_AFTER_CONNECTION).await;
                 }
+
+                // Even if connection was successful - try again after a while to update peers.
+                let duration = rand::thread_rng().gen_range(5..=DELAY_AFTER_CONNECTION);
```
ljedrz, line 59: every 5 seconds seems pretty often, doesn't it :thinking:? I wouldn't reconnect more often than once every 3 minutes

lwojtow (approved):

src/tools/crawler/crawler.rs
Rqnsom: With the old solution, where we tried to reconnect every 5 minutes (a fixed time), the number of good nodes took up to 30 minutes to reach number 50 (it doesn't go beyond that - maybe it's because of my VPN)

while with this solution, it reaches that number within 5 minutes

Rqnsom: but how I see it, the first few iterations where we scout the network should be more frequent,
then once we crawled most of the nodes multiple times, then we can slow down our crawling retries.

That would be the best approach to get reasonable results quickly and allow the crawler to run for a long time while it won't spam the network too much

Rqnsom: but I don't want to increase the complexity here. but I do like the part where we randomize the request time periods.
is it okay if we configure it to send requests every 3-5 minutes?

for my local tests, I can always tweak those values :-)

ljedrz: yeah, randomizing is a good idea, I'd just like to avoid any safeguards to trigger in the nodes we connect to :+1:

Rqnsom: > I'd just like to avoid any safeguards to trigger in the nodes we connect to

hmm, that sounds like an interesting defensive mechanism, didn't occur to me haha
(seems like it doesn't exist here though)

Done (the change)

PR #121, opened 2023-02-03
chore: crawler improvements

lwojtow:
* reduce logging
* check for handshake in background task
* make crawler rpc output optional


Rqnsom (approved):
LGTM

PR #126, opened 2023-02-23
feat(node): introduce wait_until_exit API

Rqnsom:
**`wait_until_exit` API**
A non-blocking function which periodically checks the node's status code.

**Bonus info:**
Once the async `Drop` trait support is introduced in Rust, we can remove the loop and use a non-blocking `tokio::process::Command` and then call a tokio version of `wait()` function on it.

 Because, calling `wait()` on `std::process::Command` will block the
 entire tokio runtime and the whole test process will be stopped until
 the node is actually killed - which is bad if we have other
 tokio threads running.

 So looping with a non-blocking `try_wait()` is the alternative solution.

### DIFF
--- a/src/setup/node.rs
+++ b/src/setup/node.rs
@@ -3,7 +3,7 @@ use std::{
     fs, io,
     net::{Ipv4Addr, SocketAddr, SocketAddrV4},
     path::{Path, PathBuf},
-    process::{Child, Command, Stdio},
+    process::{Child, Command, ExitStatus, Stdio},
 };
 
 use anyhow::Result;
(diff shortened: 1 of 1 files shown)


src/setup/node.rs
```diff
@@ -274,21 +274,45 @@ pub struct Node {
...
+            .map_err(|e| format!("unable to create builder: {e:?}"))
             .unwrap()
     }
 
     pub fn stop(&mut self) -> io::Result<ChildExitCode> {
         match self.child.try_wait()? {
             None => self.child.kill()?,
-            Some(code) => return Ok(ChildExitCode::ErrorCode(code.code())),
+            Some(status) => return Ok(ChildExitCode::ErrorCode(status.code())),
         }
```
Rqnsom, line 285: info for reviewers:

renamed `code` to `status` because the [try_wait](https://doc.rust-lang.org/std/process/struct.Child.html#method.try_wait) returns:
```
If the child has exited, then Ok(Some(status)) is returned. If the exit status is not available at this time then Ok(None) is returned. If an error occurs, then that error is returned.
```

also, `exit` renamed to `exit_status` because wait() function returns `ExitStatus`

lwojtow (approved):

asmie (approved):

PR #128, opened 2023-02-24
feat: rippled ci

zeapoz:
Integrates a new workflow for building and running a `rippled` instance through the test suite and logging the results daily under `results/rippled`.

~~**Note:**
The setup of the initial node state is currently relying on a long wait period (5m) before attempting to run `tools/transfer.py`. This behavior is of course not ideal and should be the subject of a future PR.~~

### DIFF
--- a/.github/workflows/rippled.yml
+++ b/.github/workflows/rippled.yml
@@ -1,11 +1,8 @@
-name: test-suite
+name: rippled
 
 on:
-  # schedule:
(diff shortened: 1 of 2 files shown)


Rqnsom (approved):
Nice work!

This can be used as a reference for our local setups

aphelionz (changes requested):
let's remove the `retry` action and simply let the CI fail. we can add it back if it becomes a real problem

Rqnsom (approved):

PR #129, opened 2023-02-24
feat(crawler): add Display trait

Rqnsom:
- formatted printout with compressed info so it doesn't clutter the output with many newlines
- removed `(super)` from `pub(super)` because the `crawl` mod will be taken out and used by other tests from this binary (source code tree)

example of the printout:
```
 server: build_version: 1.9.4, server_state: connected, server_uptime: 36
 peerlist:
[00034]: 100.24.12.73:51235                     AyT/MLgoFGxwXrV8t0WVxZFQw8LgV4ypYXn8rp5ewGzd
[out]:                          version: rippled-1.9.4                  shards: None ledgers: Some("77995485-78015485")

[00034]: 13.48.204.149:51235                    ArO1nS71bRFvZPx3qWtev20+k3D5B4Aji9k68Wgprjyu
[out]:                          version: rippled-1.9.4                  shards: None ledgers: Some("77995485-78015485")

[00034]: 13.49.231.59:51235                     AmRS+0qkiLq9fwm7y127MsHOWjMriljE3bdhFyMHxpIN
[out]:                          version: rippled-1.9.4                  shards: None ledgers: Some("77995485-78015485")

[00034]: 18.232.83.47:51235                     A7t8t5IDUoV8r5jFNZ/u+ym9nWx+5mSMnRiYuxTBkV9f
[out]:                          version: rippled-1.9.4                  shards: None ledgers: Some("77995485-78015485")

[00034]: 3.122.10.154:51235                     A/7PxSIKPdsfBuoI4u87SRup4yCK+Ric+ijvhyPmq8x7
[out]:                          version: rippled-1.9.4                  shards: None ledgers: Some("77995485-78015485")

[00034]: 34.230.16.97:51235                     AmMC5HVYfRTFKKO+O+enrnm70GE0SknM8pXrcYnV2sKa
[out]:                          version: rippled-1.9.4                  shards: None ledgers: Some("77995485-78015485")
```

### DIFF
--- a/src/tools/crawler/crawl.rs
+++ b/src/tools/crawler/crawl.rs
@@ -1,6 +1,7 @@
 //! Contains structs and methods to crawl ripple network according to instruction at https://xrpl.org/peer-crawler.html
 
 use std::{
+    fmt,
     net::{IpAddr, SocketAddr},
     time::Duration,
 };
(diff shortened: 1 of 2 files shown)


zeapoz (approved):
LGTM :+1:

src/tools/crawler/crawl.rs
```diff
@@ -10,70 +11,170 @@ use serde::Deserialize;
...
 
-#[allow(dead_code)]
-#[derive(Debug, Deserialize)]
-pub(super) struct Peer {
-    pub(super) complete_ledgers: Option<String>,
-    pub(super) complete_shards: Option<String>,
-    pub(super) ip: Option<String>,
-    pub(super) port: Option<Port>,
-    pub(super) public_key: String,
+/// Each member of the overlay.active array is an object with the following fields.
```
asmie, line 14: The dot is not in the right place.

src/tools/crawler/crawl.rs
```diff
@@ -10,70 +11,170 @@ use serde::Deserialize;
...
+    pub public_key: String,
+
+    /// Indicating whether the TCP connection to the peer is incoming or outgoing.
+    ///
+    /// The value is "in" or "out".
     #[serde(rename = "type")]
-    pub(super) connection_type: String,
-    pub(super) uptime: u32,
-    pub(super) version: String,
+    pub connection_type: String,
```
asmie, line 41: I'm not aware about the source of this information (maybe it's somewhere else also string and then it's fine to have string here), but if that field can have only two possible values wouldn't enum be more appropriate?

src/tools/crawler/crawl.rs
```diff
@@ -10,70 +11,170 @@ use serde::Deserialize;
...
+pub struct Overlay {
+    pub active: Vec<Peer>,
 }
 
+impl fmt::Display for Overlay {
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        // Sort the peers here by the IP address.
+        let mut peers = self.active.clone();
+        peers.sort_by(|a, b| {
+            let a_ip = a.ip.clone().unwrap_or_default();
```
asmie, line 105: Just wondering if cloning is here a must and something else could be used like:
`a.ip.map(|a| a.cmp(&b.ip.unwrap_or_default())).unwrap_or(Ordering::Less)`

asmie (approved):
LGTM 👍

src/tools/crawler/crawl.rs
Rqnsom: Ah, missed this one.

I like the idea, but since our `Option` contains a `String`, every `unwrap` will move out that value from the `Option` - I haven't found any quick solution to make it simpler without a `clone`

PR #130, opened 2023-03-01
refactor(synth_node): more flexible handshake cfg

Rqnsom:
TestConfig used to be an input for some conformance tests and also for a SyntheticNode creation. Now, it's decoupled:
- SyntNodeCfg ---> creation of the SyntheticNode
- TestConfig ---> used only for a test configuration (in conformance)

Also - the Handshake config has been moved to src/protocol/ and is used in both InnerNode and SyntNodeCfg.
Duplicated InnerNodeCfg handshake elements have been removed.

### DIFF
--- a/src/protocol/handshake.rs
+++ b/src/protocol/handshake.rs
@@ -24,6 +24,31 @@ enum NodeType {
     Private = 32,
 }
 
+/// Handshake configuration allows some customization of the handshake procedure.
+#[derive(Clone)]
+pub struct HandshakeCfg {
+    /// Will flip a random bit in a random byte of shared value used for session signing.
+    pub bitflip_shared_val: bool,
+
+    /// Will flip a random bit in a random byte of the public key.
+    pub bitflip_pub_key: bool,
+
+    /// Identification header to be set during a handshake.
+    /// Either 'User-Agent' or 'Server' depending on connection side.
+    pub ident: String,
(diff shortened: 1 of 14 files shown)


src/protocol/handshake.rs
```diff
@@ -24,6 +24,31 @@ enum NodeType {
...
 
+/// Handshake configuration allows some customization of the handshake procedure.
+#[derive(Clone)]
+pub struct HandshakeCfg {
+    /// Will flip a random bit in a random byte of shared value used for session signing.
+    pub bit_flip_shared_val: bool,
+
+    /// Will flip a random bit in a random byte of the public key.
+    pub bit_flip_pub_key: bool,
+
```
kylegranger, line 35: matter of taste here, but I would go with one less underscore and treat bitflip as a one unit.  I think it might improve readability later on in some of the longer lines & variable names.
```
   pub bitflip_shared_val: bool,

   /// Will flip a random bit in a random byte of the public key.
   pub bitflip_pub_key: bool,
```

asmie (approved):
LGTM

kylegranger (approved):
LGTM 👍

src/protocol/handshake.rs
Rqnsom: nice, I like it! thanks.
Done

PR #131, opened 2023-03-01
Introduce setup script

zeapoz:
_Introduces a convenience shell script for setting up the work environment._

- The script performs the following:
  - Sets up `~/.ziggurat/ripple` directories
  - Creates a default config file
  - ~~Adds IP addresses used in performance tests~~
  - Sets up an initial node state

- Documentation changes are included.

### DIFF
--- a/README.md
+++ b/README.md
@@ -15,7 +15,16 @@ Ziggurat is written in stable Rust; you can install the Rust toolchain by follow
 3. Make sure you have python3 installed. You should be able to run `python3 --version`.
 4. Install `xrpl` python lib: `pip3 install xrpl-py`.
 
-   ##### Important note!
+   ##### **Mandatory step for Mac users!**
+   On MacOS, the `coreutils` package is also required. It can be installed with: `brew install coreutils`
+
+   Also make sure that these two `127.0.0.x` (where `x != 1`) addresses are enabled:
+   ```bash
+   sudo ifconfig lo0 alias 127.0.0.2 up;
+   sudo ifconfig lo0 alias 127.0.0.3 up;
+   ```
+
+   ##### **Important note!**
    The `xrlp` library depends on legacy openSSL functions that are disabled by default. In the case of error, make sure to explicitly enable legacy functions in the config file, like so:
    ```
    openssl_conf = openssl_init
@@ -66,11 +75,6 @@ python3 ./tools/ips.py --subnet 1.1.1.0/24 --file src/tools/ips.rs --dev_prefix 
 Add the whole subnet to the loopback device - can also be used on Linux (device name - Linux: `lo`, MacOS: `lo0`):
 ```bash
 python3 ./tools/ips.py --subnet 1.1.0.0/24 --file src/tools/ips.rs --dev lo0
-```
-On MacOS, make sure these two `127.0.0.x` (where `x != 1`) addresses are enabled:
-```bash
-sudo ifconfig lo0 alias 127.0.0.2 up;
-sudo ifconfig lo0 alias 127.0.0.3 up;
 ```
 
 Read ./tools/ips.py for more details.
--- a/README.md
+++ b/README.md
@@ -2,44 +2,21 @@
 
 The Ziggurat implementation for XRPLF's `rippled` nodes.
 
+## Prerequisites
+Ziggurat is written in stable Rust; you can install the Rust toolchain by following the official instructions [here](https://www.rust-lang.org/learn/get-started)
+
 ## Getting started
 
+### Preconditions
 1. Clone this repository.
 2. Build [rippled](https://github.com/XRPLF/rippled) from source.
-3. Create the `~/.ziggurat/ripple/setup` directories, and copy the `setup/validators.txt` file there.
-   ```
-   cp setup/validators.txt ~/.ziggurat/ripple/setup
-   ```
-4. In the same directory create a `config.toml` with the following contents:
-   ```
-   path = "<path to the directory where you built rippled>"
-   start_command = "./rippled"
-   ```
-5. Create a package of IP addresses which are required for performance tests. From the root repository directory run, e.g.:
-   Under Linux (to generate dummy devices with addresses):
-   ```
-   sudo python3 ./tools/ips.py --subnet 1.1.1.0/24 --file src/tools/ips.rs --dev_prefix test_zeth
-   ```
-   Under MacOS or Linux (to add whole subnet to loopback device - under Linux: lo, MacOS: lo0):
-   ```
-   sudo python3 ./tools/ips.py --subnet 1.1.1.0/24 --file src/tools/ips.rs --dev lo0
-   ```
-   Read ./tools/ips.py for more details.
-6. Run tests with:
-   ```
-   cargo +stable t -- --test-threads=1
-   ```
 
-### Initial state
-Specific tests require an initial node state to be set up.
-Follow the steps below to save an initial state that can be loaded later in certain tests.
+#### Running setup script
+3. Make sure you have python3 installed. You should be able to run `python3 --version`.
+4. Install `xrpl` python lib: `pip3 install xrpl-py`.
 
-#### Preparation (needs to be done once)
-1. Make sure you have python3 installed. You should be able to run `python3 --version`.
-2. Install `xrpl` python lib: `pip3 install xrpl-py`.
-
-##### Important note!
-The `xrlp` library depends on legacy openSSL functions that are disabled by default. In the case of error, make sure to explicitly enable legacy functions in the config file like so:
+   ##### Important note!
+   The `xrlp` library depends on legacy openSSL functions that are disabled by default. In the case of error, make sure to explicitly enable legacy functions in the config file, like so:
    ```
    openssl_conf = openssl_init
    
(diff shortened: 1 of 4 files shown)


tools/setup_env.sh
```diff
@@ -0,0 +1,114 @@
...
+    # Print file contents so the user can check whether the path is correct
+    cat $ZIGGURAT_RIPPLED_SETUP_CFG_FILE
+    echo
+}
+
+setup_ip_addresses() {
+    echo "--- Creating a package of IP addresses required for performance tests"
+    sudo python3 ./tools/ips.py --subnet 1.1.1.0/24 --file src/tools/ips.rs --dev_prefix test_zeth
+    if [ $(uname) == "Linux" ]; then
+        sudo python3 ./tools/ips.py --subnet 1.1.1.0/24 --file src/tools/ips.rs --dev lo
```
asmie, line 37: Does it mean that on Linux it `ips.py` will be executed twice? First time on line 35 and second time here resulting in having the same subnet added to test_zeth devices and to lo device?

tools/setup_env.sh
```diff
@@ -0,0 +1,114 @@
...
+    echo "start_command = \"./$RIPPLED_BIN_NAME --silent\"" >> $ZIGGURAT_RIPPLED_SETUP_CFG_FILE
+
+    # Print file contents so the user can check whether the path is correct
+    cat $ZIGGURAT_RIPPLED_SETUP_CFG_FILE
+    echo
+}
+
+setup_ip_addresses() {
+    echo "--- Creating a package of IP addresses required for performance tests"
+    sudo python3 ./tools/ips.py --subnet 1.1.1.0/24 --file src/tools/ips.rs --dev_prefix test_zeth
```
asmie, line 35: One more thing - maybe subnet and device should be also some var configured on the top of the script to easily be changed by the user?

tools/setup_env.sh
zeapoz: Ah yes that's true, good catch! :+1:

asmie: So the same would be on Mac - first time on 35 and second time on line 39.

tools/setup_env.sh
zeapoz: I wonder if we maybe should use `lo` and `lo0` as default devices, and then by precedence of some exported variable allow them to be overridden instead?

asmie: They could be default devices. I think that most users will choose to add addresses to the loopback device and it could be rarely used case when someone would like to add it to other device or create new device for each address.

tools/setup_env.sh
```diff
@@ -0,0 +1,114 @@
...
+if [ "`basename $REPO_ROOT`" != "xrpl" ]; then
+    # Wrong root directory, check for rename compared to origin url.
+    ORIGIN_URL=$(git config --local remote.origin.url|sed -n 's#.*/\([^.]*\)\.git#\1#p')
+    if [ "$ORIGIN_URL" != "xrpl" ]; then
+        echo "Aborting. Use this script only from the ziggurat/xrpl repo."
+        exit 1
+    fi
+fi
+
+# Setup the main ziggurat directory in the home directory
```
asmie, line 101: Probably `rm -rf $ZIGGURAT_RIPPLED_DIR` does not mean "Setup the main ziggurat directory" :)

asmie (changes requested):
Overall LGTM.
There is a need to change execution of ips.py script twice on Mac and Linux.

tools/setup_env.sh
```diff
@@ -0,0 +1,114 @@
...
+    kill -2 $(pidof cargo)
+    echo "--- Performing cleanup"
+    # Remove unneeded and temporary files
+    rm $ZIGGURAT_RIPPLED_STATEFUL_DIR/*/rippled.cfg
+    rm -rf $ZIGGURAT_RIPPLED_TESTNET_DIR
+    echo
+}
+
+# Verify the repo location
+if [ "$(git rev-parse --is-inside-work-tree 2>/dev/null)" != "true" ]; then
```
asmie, line 87: Very nice 👍

asmie (approved):

Rqnsom (approved):
LGTM, let me try it on Mac before, I'll notify you for the resutls

tools/setup_env.sh
```diff
@@ -0,0 +1,114 @@
...
+
+setup_config_file() {
+    echo "--- Setting up configuration file"
+    echo "Creating $ZIGGURAT_RIPPLED_SETUP_CFG_FILE with contents:"
+    mkdir -p $ZIGGURAT_RIPPLED_SETUP_DIR
+    echo
+    echo "# Rippled installation path" > $ZIGGURAT_RIPPLED_SETUP_CFG_FILE
+    echo "path = \"$RIPPLED_BIN_PATH\"" >> $ZIGGURAT_RIPPLED_SETUP_CFG_FILE
+    echo "# Start command with possible arguments" >> $ZIGGURAT_RIPPLED_SETUP_CFG_FILE
+    echo "start_command = \"./$RIPPLED_BIN_NAME --silent\"" >> $ZIGGURAT_RIPPLED_SETUP_CFG_FILE
```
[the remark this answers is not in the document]
Rqnsom: yeah, let's remove it from here. I was debugging something and I couldn't see any logs then figured this option was blocking my node output

we can rather set `log_to_stdout` to false in the testnet test (`src/setup/testnet/mod.rs`)

Rqnsom (approved):
once the `gtimeout` is used from coreutils, the script works on Mac, but

this thing from the old Readme file was also missing - after I applied it manually, everything worked:
```
    sudo ifconfig lo0 alias 127.0.0.2 up;
    sudo ifconfig lo0 alias 127.0.0.3 up;
```
but let's do it for now, let's make this a mandatory step before running the script with a note that it's mandatory for Mac users
so we can submit this PR


------- unrelated comment / possible improvement ------

but! I'm not fond of that (Mac mandatory) step at all. we should try to avoid putting these requirements on our user - plus the user interface for the setup would be cleaner without it - which makes everybody happy

so I started checking if we could avoid doing this maybe:
I managed to set up multiple rippled instances on the same localhost IP with different ports - which means we can run parallel rippled instances simultaneously. that implies we could run conf/resist tests without the single thread requirement, which is nice

the trouble here is the RPC endpoint which uses the fixed `5005` port.
which is, well, tricky since once we start using a dynamic RPC address, we will have to fetch it from rippled.cfg - and we use this address in all py scripts + and in the test suite
complex but doable

but if we change this, we would simplify the stateful node setup in the `node.rs`
and also - we could easily run all tests in parallel by default

README.md
```diff
@@ -2,44 +2,21 @@
...
-### Initial state
-Specific tests require an initial node state to be set up.
-Follow the steps below to save an initial state that can be loaded later in certain tests.
 
-#### Preparation (needs to be done once)
-1. Make sure you have python3 installed. You should be able to run `python3 --version`.
-2. Install `xrpl` python lib: `pip3 install xrpl-py`.
+#### Running setup script
+3. Make sure you have python3 installed. You should be able to run `python3 --version`.
+4. Install `xrpl` python lib: `pip3 install xrpl-py`.
```
Rqnsom, line 16: `brew install coreutils `

for Mac only

tools/setup_env.sh
```diff
@@ -0,0 +1,105 @@
...
+    cargo t setup::testnet::test::run_testnet -- --ignored &
+    echo
+    sleep $ACCOUNT_QUERY_DELAY_MIN
+    echo "--- Querying account info"
+    # Run account query until it responds with "ResponseStatus.SUCCESS" or MAX_ATTEMPTS is reached
+    ACCOUNT_QUERY_TIMEOUT_SEC=5s
+    ACCOUNT_QUERY_MAX_ATTEMPTS=5
+    ACCOUNT_QUERY_NUM_ATTEMPTS=0
+    until [ $ACCOUNT_QUERY_NUM_ATTEMPTS -gt $(($ACCOUNT_QUERY_MAX_ATTEMPTS-1)) ] \
+        || timeout $ACCOUNT_QUERY_TIMEOUT_SEC python3 tools/account_info.py | grep "ResponseStatus.SUCCESS"; do
```
Rqnsom, line 52: for Mac we should use the `gtimeout` cmd here

Rqnsom (approved):

tools/setup_env.sh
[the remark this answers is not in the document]
zeapoz: Done :white_check_mark:

tools/setup_env.sh
```diff
@@ -0,0 +1,105 @@
...
+    fi
+    echo "Established a connection with the genesis account"
+    echo
+    echo "--- Executing transfer script"
+    python3 tools/transfer.py
+    # Copy the node's files to directory referenced by constant pub const STATEFUL_NODES_DIR
+    cp -a $ZIGGURAT_RIPPLED_TESTNET_DIR $ZIGGURAT_RIPPLED_STATEFUL_DIR
+    echo
+    echo "--- Gracefully stopping the network"
+    kill -2 $(pidof cargo)
```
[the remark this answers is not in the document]
zeapoz: Done :white_check_mark:

tools/setup_env.sh
```diff
@@ -0,0 +1,105 @@
...
+    echo "path = \"$RIPPLED_BIN_PATH\"" >> $ZIGGURAT_RIPPLED_SETUP_CFG_FILE
+    echo "# Start command with possible arguments" >> $ZIGGURAT_RIPPLED_SETUP_CFG_FILE
+    echo "start_command = \"./$RIPPLED_BIN_NAME --silent\"" >> $ZIGGURAT_RIPPLED_SETUP_CFG_FILE
+
+    # Print file contents so the user can check whether the path is correct
+    cat $ZIGGURAT_RIPPLED_SETUP_CFG_FILE
+    echo
+}
+
+setup_initial_node_state() {
```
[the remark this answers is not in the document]
zeapoz: Done :white_check_mark:

README.md
zeapoz: Done :white_check_mark:

tools/setup_env.sh
zeapoz: Done :white_check_mark:

PR #132, opened 2023-03-02
Migrate crawler from IpAddr to SocketAddr & NodesIndices.

kylegranger:
As we did for zcash, I've updated the XRPL crawler to use the latest ziggurat-core objects.
- `IpAddr` has been changed to `SocketAddr` throughout.
- since the port is often an `Option` variable, use a reasonable default when creating a `SocketAddr` instances from an `IpAddr`
- an additional JSON-RPC method has been added to support writing a network summary to disk directly (rather than doing it from the current `getmetrics` call.  The crawler readme has been correspondingly updated.  It solves a serialization problem, as well as providing a cleaner API.
- When data is written to disk directly, it gets wrapped in a JSON-RPC response.  This way, all files are consistent.


src/tools/crawler/rpc.rs
```diff
@@ -37,18 +37,26 @@ pub async fn initialize_rpc_server(rpc_addr: SocketAddr, rpc_context: RpcContext
...
                 // TODO: consider some checks against directory traversal
-                if let Err(e) = fs::write(path, content) {
+                if let Err(e) = fs::write(path, response) {
                     warn!("Unable to write to file: {}", e);
                 }
-                Ok(RpcOutput::Length(length))
+                Ok(length)
             } else {
-                Ok(RpcOutput::Summary(rpc_context.0.lock().unwrap().clone()))
+                Ok(-1)
```
asmie, line 59: Just a little comment could be useful what does would mean to the caller - is this an error code with some meaning or just some random negative to indicate there was an error (path was not found)?

src/tools/crawler/rpc.rs
```diff
@@ -37,18 +37,26 @@ pub async fn initialize_rpc_server(rpc_addr: SocketAddr, rpc_context: RpcContext
...
                 let content = serde_json::to_string(rpc_context.0.lock().unwrap().deref())?;
-                let length = content.len();
+                // Wrap our NetworkSummary in a JSON-RPC response envelope
+                let response =
+                    "{\"jsonrpc\":\"2.0\",\"result\":".to_owned() + &content + ",\"id\":0}";
+                let length = response.len() as i32;
                 // TODO: consider some checks against directory traversal
-                if let Err(e) = fs::write(path, content) {
+                if let Err(e) = fs::write(path, response) {
                     warn!("Unable to write to file: {}", e);
```
asmie, line 55: If path is not right this function returns -1, but if path is ok but cannot write to file we'll return the `length` just as if writing was successful - is that correct?

asmie (approved):
LGTM 👍

src/tools/crawler/rpc.rs
kylegranger: yes, this was completely naked here -- I've just checked in a change where an object get returned in the result: with status (success/fail), length, and message.

src/tools/crawler/rpc.rs
kylegranger: good find, yes, it's a third return branch by itself.

src/tools/crawler/rpc.rs
kylegranger: // expected response
```
{
    "jsonrpc": "2.0",
    "result": {
        "status": "success",
        "length": 12400,
        "message": ""
    },
    "id": 0
}
```
// no file parameter
```
{
    "jsonrpc": "2.0",
    "result": {
        "status": "fail",
        "length": -1,
        "message": "No file parameter in params"
    },
    "id": 0
}
```
// bad file path passed in, or other file I/O error
```
{
    "jsonrpc": "2.0",
    "result": {
        "status": "fail",
        "length": -1,
        "message": "Unable to write file: No such file or directory (os error 2)"
    },
    "id": 0
}
```

PR #135, opened 2023-03-07
feat(handshake): resistance test - fuzz fields pt1

Rqnsom:


src/tests/resistance/handshake.rs
```diff
@@ -101,6 +111,138 @@ async fn r001_t2_HANDSHAKE_reject_if_server_too_long() {
...
+    // in the field.
+    let cfg = gen_cfg(gen_huge_string(WS_HTTP_HEADER_MAX_SIZE));
+    assert!(!run_handshake_req_test_with_cfg(cfg, debug).await);
+
+    // Use a huge value which the node will always reject.
+    let cfg = gen_cfg(gen_huge_string(WS_HTTP_HEADER_INVALID_SIZE));
+    assert!(!run_handshake_req_test_with_cfg(cfg, debug).await);
+
+    // Send an empty field.
+    // NOTE: hmm, should the node allow an empty name field?
```
Rqnsom, line 241: oops, let's rm this

Rqnsom: Don

src/tests/resistance/handshake.rs
```diff
@@ -101,6 +111,137 @@ async fn r001_t2_HANDSHAKE_reject_if_server_too_long() {
     node.stop().unwrap();
 }
 
+/// Decide whether to enable node logs and tracing for synthetic nodes.
+#[derive(Clone, Copy)]
+enum Debug {
```
lwojtow, line 116: I'm slightly worried about the name here.

lwojtow (approved):

src/tests/resistance/handshake.rs
Rqnsom: my logic was:
when some test is failing -> I need to debug that test which implies enabling logs/tracing
so I kinda put it under the `Debug On/Off` enum

what name would you prefer or have in mind?

src/tests/resistance/handshake.rs
```diff
@@ -101,6 +111,137 @@ async fn r001_t2_HANDSHAKE_reject_if_server_too_long() {
...
+        Self::On
+    }
+
+    fn disable() -> Self {
+        // We should use something like synth_node::disable_tracing here (still unimplemented),
+        // but we'll never use it anyway so this is good enough
+        Self::Off
+    }
+
+    /// Convert to boolean value.
```
Rqnsom, line 135:
```suggestion
    /// Convert to a boolean value.
```

src/tests/resistance/handshake.rs
```diff
@@ -101,6 +111,137 @@ async fn r001_t2_HANDSHAKE_reject_if_server_too_long() {
...
+    let handshake_established = if synthetic_node.connect(node.addr()).await.is_err() {
+        false
+    } else {
+        // Wait for any message.
+        synthetic_node
+            .expect_message(&|m: &BinaryMessage| matches!(&m, _))
+            .await
+    };
+
+    if debug.is_on() {
```
Rqnsom, line 168: no need to wait in case of successful handshake
```suggestion
    if debug.is_on() && !handshake_established {
```

src/tests/resistance/handshake.rs
```diff
@@ -101,6 +111,137 @@ async fn r001_t2_HANDSHAKE_reject_if_server_too_long() {
...
+    let cfg = gen_cfg(gen_huge_string(WS_HTTP_HEADER_MAX_SIZE));
+    assert!(!run_handshake_req_test_with_cfg(cfg, debug).await);
+
+    // Use a huge value which the node will always reject.
+    let cfg = gen_cfg(gen_huge_string(WS_HTTP_HEADER_INVALID_SIZE));
+    assert!(!run_handshake_req_test_with_cfg(cfg, debug).await);
+
+    // Send an empty field.
+    let cfg = gen_cfg(String::new());
+    assert!(!run_handshake_req_test_with_cfg(cfg, debug).await);
```
Rqnsom, line 242: it would be more visible if we use `assert_eq(false, ...)` that we expect these tests to fail
small `!` is just a tiny character here 🤔

Rqnsom (commented):
some extra self review

src/tests/resistance/handshake.rs
Rqnsom: Done

src/tests/resistance/handshake.rs
Rqnsom: Done

src/tests/resistance/handshake.rs
Rqnsom: Done - it looks more readable, but kinda strange -> let's keep it for the code readability reasons

PR #138, opened 2023-03-08
feat: crawler workflow

zeapoz:
This PR introduces a new workflow that will:

1. Compile a `rippled` node from source
2. Crawl the network, using the built node as the entry point
3. Produce network metrics under the `results/crawler` directory.


kylegranger (approved):
jawohl! 👍

PR #140, opened 2023-03-09
Feat/fuzzing http fields

lwojtow:


src/tests/resistance/handshake.rs
```diff
@@ -25,6 +28,9 @@ use crate::{
 const WS_HTTP_HEADER_MAX_SIZE: usize = 7700;
 const WS_HTTP_HEADER_INVALID_SIZE: usize = WS_HTTP_HEADER_MAX_SIZE + 300;
 
+// Number of seconds between unix and ripple epoch.
+const RIPPLE_EPOCH_OFFSET: u64 = 946684800;
```
kylegranger, line 32: I'll just mention, I used this same number when doing the validator test.
https://github.com/runziggurat/xrpl/blob/main/src/tests/conformance/post_handshake/validators.rs#L23
Maybe something to think about, how to have it exist in only one place.

kylegranger (approved):
👍

src/tests/resistance/handshake.rs
lwojtow: Let's move it some constant in one of future PR?

src/tests/resistance/handshake.rs
```diff
@@ -290,6 +331,141 @@ async fn r001_t4_HANDSHAKE_connect_as_field() {
...
+        }),
+        ..Default::default()
+    };
+
+    // Valid value for Upgrade
+    let cfg = gen_cfg("XRPL/2.2".to_owned());
+    assert!(run_handshake_req_test_with_cfg(cfg, debug).await);
+
+    // Find the largest instance value which the node could accept.
+    let cfg = gen_cfg(format!("{:#7700}", "XRPL/2.2"));
```
Rqnsom, line 444: does this repeat the "XRPL/2.2" string 7700 times, or how does the string look like?

because  I was gonna suggest trying "XRPL/2.2,XRPL/2.2"

src/tests/resistance/handshake.rs
```diff
@@ -290,6 +331,141 @@ async fn r001_t4_HANDSHAKE_connect_as_field() {
...
+    // Send correct value but prepend too many zeros.
+    let cfg = gen_cfg(format!("{:#08000}", time_now));
+    assert!(!run_handshake_req_test_with_cfg(cfg, debug).await);
+}
+
+#[allow(non_snake_case)]
+#[tokio::test]
+async fn r001_t8_HANDSHAKE_upgrade_req_field() {
+    // ZG-RESISTANCE-001
+    // Expected valid value for the "Upgrade" field in the handshake should be valid.
```
Rqnsom, line 427: a valid what? 😃

Rqnsom (commented):
just a few extra comments, we can fix those in the future PR

LGTM

PR #142, opened 2023-03-10
Feat/more handshake resistance tests

lwojtow:


src/tests/resistance/handshake.rs
```diff
@@ -466,6 +464,140 @@ async fn r001_t8_HANDSHAKE_upgrade_req_field() {
...
+        handshake: Some(HandshakeCfg {
+            http_unexpected_extra_field_and_value: Some(value),
+            ..Default::default()
+        }),
+        ..Default::default()
+    };
+
+    // "Valid" fields
+    let cfg = gen_cfg("Exec: /bin/bash".to_owned());
+    assert!(run_handshake_req_test_with_cfg(cfg, debug).await);
```
Rqnsom, line 588: let's try to repeat one Header field here, like the Upgrade field or ident field

so the same header appears twice or something.

Rqnsom (commented):
LGTM

src/tests/resistance/handshake.rs
lwojtow: I've changed this test to give "Connect-As" header with different value (comparing to the one that's always present) and surprisingly it passes. I guess we can report it in some way to the client.

SPEC.md
```diff
@@ -378,7 +378,7 @@ The test index makes use of symbolic language in describing connection and messa
 
 ### ZG-RESISTANCE-003
 
-    The node rejects the handshake when there is an in any of handshake field: 
+    The node rejects the handshake when there is error an in any of handshake field:
```
Rqnsom, line 381:
```suggestion
    The node rejects the handshake when there is an error in any of the handshake fields: 
```

src/tests/resistance/handshake.rs
Rqnsom: Nice,
but can we also keep the previous subtests:
```rust
    // "Valid" fields
    let cfg = gen_cfg("Exec: /bin/bash".to_owned());
    assert!(run_handshake_req_test_with_cfg(cfg, debug).await);

    let cfg = gen_cfg("Exec: ".to_owned());
    assert!(run_handshake_req_test_with_cfg(cfg, debug).await);

    let cfg = gen_cfg(format!("{:#7000}", "Exec: /bin/bash"));
    assert!(run_handshake_req_test_with_cfg(cfg, debug).await);

    // Invalid value (too long)
    let cfg = gen_cfg(format!("{:#9000}", "Exec: /bin/bash"));
```

asmie (approved):
👍

Rqnsom (approved):
Since @lwojtow is not available at the moment, let's merge, I can apply my suggestions in another PR

PR #143, opened 2023-03-11
feat(dev): add ping-pong mechanism dev tests

Rqnsom:
The synthetic node should automatically reply to ping messages in order to keep the connection alive.


asmie (approved):
👍

PR #145, opened 2023-03-14
feat(crawler): introduce rate limiting

zeapoz:
Our current workflow for the crawler produces no `good_nodes` on its own. This PR fixes this problem by introducing a rate limiter, only allowing `25` requests per second.

Changes:
- Adds a new dependency, `governor`.
- Introduces the `Limiter` struct, a wrapper around `governor::RateLimiter` along with `governor::Jitter` (to prevent thundering herd problem).
- Awaits `Limiter::until_ready` before sending crawl requests, to not need send more than `MAX_REQUESTS_PER_SEC` requests per second.


src/tools/crawler/main.rs
```diff
@@ -43,6 +52,30 @@ fn start_logger(default_level: LevelFilter) {
         .init();
 }
 
+struct Limiter {
+    limiter: RateLimiter<NotKeyed, InMemoryState, QuantaClock, NoOpMiddleware<QuantaInstant>>,
```
asmie, line 56: Just for fun question: have you checked which one is faster here: `QuantaClock` or `QuantaUpkeepClock`? 
As the one you've chosen is synchronous and the other one use background thread to update the clock I'm curious which one could be faster (well, we've only 25 conns right now so probably there shouldn't be any difference).

asmie (approved):
👍

src/tools/crawler/main.rs
zeapoz: No I have not, though, as you said, there probably wouldn't be a significant difference due to our low limit. There's also the added problem of `QuantaUpkeepClock` not implementing `ReasonablyRealtime`, which is a requirement for use in `until_ready`/`until_ready_with_jitter`. So, that'd have to be taken care of as well.

PR #147, opened 2023-03-16
Reuse workflows

zeapoz:
This PR:
- Moves shared `install-rippled` code into a new workflow.
- Adjusts the rippled workflow to call the `process-results` workflow under `runziggurat/zcash`.


asmie (approved):
👍

PR #148, opened 2023-03-21
chore: call new workflows

zeapoz:
- Refactors `install-* -> build-*`.
- Call `process-results` workflow from crawler
- Call `diff-with-previous` workflow from `rippled`


kylegranger (approved):
👍

PR #149, opened 2023-03-22
chore(dev): link crawl.rs with the CPR suite

Rqnsom:
- dev tests in `idle_node_in_the_background.rs` use a crawler mechanism for easier analysis of the node behavior. For that reason existing crawl.rs is reused from the crawler binary.

  A quick solution is used here - instead of making a clean copy of the file, or even moving it out from the binary directory, we are just reusing it with a symlink. In future Ziggurat projects, this shall be handled in a different fasion.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -9,6 +9,7 @@ anyhow = "1.0"
 async-trait = "0.1"
 base64 = "0.21.0"
 bytes = "1"
+chrono = "0.4"
 fs_extra = "1.2"
 governor = "0.5.1"
 hex = "0.4"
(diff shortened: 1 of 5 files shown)


kylegranger (approved):
LGTM

Rqnsom (commented):
> what are alternatives? 

I'll create a thread on Slack for this later today so we can chat about it

> Are they harder to implement or they can take much more time to be incorporated

This! I wouldn't say harder to implement, I would say it's too late and too much effort for this specific project since we are finalizing work here.

PR #150, opened 2023-03-22
feat: add synth_node_bin files

Rqnsom:
A synthetic node binary can be used to interact with the node in the background from a different runtime environment.

### DIFF
--- a/synth_node_bin/Cargo.toml
+++ b/synth_node_bin/Cargo.toml
@@ -1,0 +1,16 @@
+[package]
+name = "synth_node_bin"
+version = "0.1.0"
+edition = "2021"
+
+[dependencies]
+anyhow = "1.0"
+clap = { version = "4.1", features = ["derive"] }
+rand = "0.8"
+tokio = "1.25"
+tracing-subscriber = "0.3"
+ziggurat-xrpl = { path = "../" }
+
+[dependencies.tracing]
+version = "0.1"
+default-features = false
--- a/synth_node_bin/src/main.rs
+++ b/synth_node_bin/src/main.rs
@@ -1,0 +1,129 @@
+//! A synthetic node binary can be used to interact with the node in the
+//! background from a different runtime environment.
+//!
+//! This is a quick solution - in future Ziggurat projects it will be done differently.
+//!
+//! Remaining worklist (probably not going to be implemented in this repo):
+//!   - Add an argument option to choose specific action in order to support
+//!     different synthetic node binary implementations at once. A few examples:
+//!     ```
+//!        ./synthetic_node_bin --action=A    // Runs an idle/friendly synthetic node
+//!        ./synthetic_node_bin --action=B    // Runs a wild synthetic node which does something funny
+//!     ```
+use std::{net::SocketAddr, process::ExitCode};
+
+use anyhow::Result;
+use clap::Parser;
+use tokio::time::{sleep, Duration};
+use ziggurat_xrpl::tools::{config::SynthNodeCfg, synth_node::SyntheticNode};
+
+/// A synthetic node which can connect to the XRPL node and preform some
+/// actions independently.
+#[derive(Parser)]
+#[command(author, version, about, long_about = None)]
+struct CmdArgs {
+    /// An address of the node in the <ip>:<port> format.
+    #[arg(short = 'i', long)]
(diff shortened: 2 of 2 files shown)


synth_node_bin/Cargo.toml
```diff
@@ -0,0 +1,16 @@
...
+name = "synth_node_bin"
+version = "0.1.0"
+edition = "2021"
+
+[dependencies]
+anyhow = "1.0"
+clap = { version = "4.1", features = ["derive"] }
+rand = "0.8"
+tokio = "1.25"
+tracing-subscriber = "0.2.0"
```
asmie, line 11: Is it that crate: https://crates.io/crates/tracing-subscriber ?
It's version is v0.3.16.

synth_node_bin/src/main.rs
```diff
@@ -0,0 +1,128 @@
...
+use anyhow::Result;
+
+/// A synthetic node which can connect to the XRPL node and preform some
+/// actions independently.
+#[derive(Parser)]
+#[command(author, version, about, long_about = None)]
+struct CmdArgs {
+    /// An address of the node in the <ip>:<port> format.
+    #[arg(short = 'i', long)]
+    node_ip: Option<SocketAddr>,
```
asmie, line 27: Maybe `node_addr` as it's not `IpAddr` type but `SocketAddr` so with port.

synth_node_bin/src/main.rs
```diff
@@ -0,0 +1,128 @@
...
+    /// Enable tracing.
+    #[arg(short = 't', long, default_value_t = false)]
+    tracing: bool,
+}
+
+#[tokio::main]
+async fn main() -> ExitCode {
+    let args = CmdArgs::parse();
+
+    let node_addr = if let Some(addr) = args.node_ip {
```
asmie, line 42: Maybe it would be shorter:
```
    let node_addr = args.node_ip.unwrap_or_else(|| {
        eprintln!("Node address should be provided.");
        return ExitCode::FAILURE;
    });
```

synth_node_bin/src/main.rs
```diff
@@ -0,0 +1,128 @@
...
+    s
+}
+
+// --- TRY NOT TO CHANGE THE ABOVE CODE ---
+
+fn configure_synth_node() -> SynthNodeCfg {
+    let mut cfg = SynthNodeCfg::default();
+
+    cfg.handshake = cfg.handshake.map(|mut hs| {
+        hs.http_ident = gen_random_str(19);
```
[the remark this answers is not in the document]
Rqnsom: good point! it was rather randomly chosen, there is no hidden meaning behind the number

no need for `const` since we use it only once, and now we will also have a comment

synth_node_bin/src/main.rs
Rqnsom: I like the idea, but we can't use the `return` statement in closure to return the code from the `main` fn.

I'm not sure if there is some sort of alternative solution out there yet for this particular situation - to make it simpler (like your code snippet)

synth_node_bin/Cargo.toml
asmie: I'm looking sometimes for packages to check what they do :) 
But we can add to readme:
`[dependency status](https://deps.rs/repo/github/runziggurat/xrpl)`
Just to have status everytime we open the page 😃

synth_node_bin/src/main.rs
Rqnsom: Done

synth_node_bin/Cargo.toml
Rqnsom: I'll specify without the `patch` part, it should always be backwards compatible that way

Done

synth_node_bin/src/main.rs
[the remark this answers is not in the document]
Rqnsom: Done

synth_node_bin/Cargo.toml
Rqnsom: > I'm looking sometimes for packages to check what they do :)
But we can add to readme:
[dependency status](https://deps.rs/repo/github/runziggurat/xrpl)
Just to have status everytime we open the page smiley

yes, please! :D
will do that in the next PR

PR #153, opened 2023-03-27
chore: add workflow badge, move both badges below title

kylegranger:
extension of the Improvement Bucket Task: `Add deps.rs badges to all our readme documents`
- added workflow CI badge
- moved badges below title


zeapoz (approved):
LGTM 👍

PR #156, opened 2023-04-04
New ips.py script and node list handling

kylegranger:
This PR refactors the way the ips.py script is gotten, and how the list of nodes is handled.  It addresses the following issues in the improvement bucket:
- How to share common scripts across different repos
- Think how to avoid having modified file in the repo after running the script
- Move the ips.py script to some common place for scripts in ziggurat-core

Main features:
- script to fetch `ips.py `from `ziggurat-core` (documented in the readme)
- replaced static IPS list in the generated `ips.rs` with reading a json file (`tools/ips_list.json`) and passing back an array to the callers in the test/performance modules: connections, get_trans, ping_pong.

I suppose there are 'many ways to Rome' for something like this.  I'm not married to this particular strategy, if somebody has a better idea.


tools/core_scripts.sh
zeapoz, line 1: Do we really need a new file here? Maybe it's better to just have the command in the `README.md` instead

src/tools/ips.rs
```diff
@@ -8,4 +8,31 @@
...
+
+const IPS_LIST_PATH: &str = "./tools/ips_list.json";
+
+#[derive(Default, Clone, Deserialize, Debug)]
+struct IpsList {
+    pub nodes: Vec<String>,
+}
+
+/// Read in the node array generated by the ips.py
+/// script.  If file does not exist, return an empty
```
zeapoz, line 23:
```suggestion
/// script. If file does not exist, return an empty
```

src/tools/ips.rs
```diff
@@ -8,4 +8,31 @@
...
+/// array.
+fn load_ips_nodes(filepath: &str) -> Vec<String> {
+    let result = fs::read_to_string(filepath);
+    if result.is_err() {
+        println!("no file {filepath}");
+        return Vec::<String>::new();
+    }
+    let jstring = result.unwrap();
+    let ips_list: IpsList = serde_json::from_str(&jstring).unwrap();
+    ips_list.nodes
```
zeapoz, line 33: Probably could use a `match` statement here.

zeapoz (commented):
Some nits, other than that LGTM :+1:

tools/core_scripts.sh
kylegranger: yeah, good idea.  I was in the dialectic of submodule vs. wget-script, didn't see the forest for the trees. 😄

src/tools/ips.rs
```diff
@@ -8,4 +8,31 @@
...
+
+/// Read in the node array generated by the ips.py
+/// script.  If file does not exist, return an empty
+/// array.
+fn load_ips_nodes(filepath: &str) -> Vec<String> {
+    let result = fs::read_to_string(filepath);
+    if result.is_err() {
+        println!("no file {filepath}");
+        return Vec::<String>::new();
+    }
```
Rqnsom, line 30: hm, let's rather fail (panic with `expect`) here directly, no need to be nice 😃

if the file is not found, maybe leave a message to the user to check if he needs to run our script in our Readme documentaiton

src/tools/ips.rs
```diff
@@ -8,4 +8,31 @@
...
+    let jstring = result.unwrap();
+    let ips_list: IpsList = serde_json::from_str(&jstring).unwrap();
+    ips_list.nodes
+}
+
+pub fn ips() -> Vec<String> {
```
Rqnsom, line 36: Let's leave a short fn comment with `///` since it's a public function

tools/core_scripts.sh
kylegranger: done

src/tools/ips.rs
kylegranger: done

README.md
```diff
@@ -62,6 +62,12 @@ cargo +stable t -- --test-threads=1
 ### Run performance tests
 Create a package of IP addresses which are required for performance tests.
 
+You must first fetch the `ips.py` script from the ziggurat-core repository.  Run this:
+
+```bash
+./tools/core_scripts.sh
+```
+
```
Rqnsom: we could even remove this step and hide this part from the user by calling below `core_scripts.sh` script instead of `ips.py` (make that script a wrapper around `ips.py`):
```
./tools/core_scripts.sh --subnet 1.1.1.0/24 --file src/tools/ips.rs --dev_prefix test_zeth
```

and in the script itself, we would just forward these arguments to the real `ips.py` script call.

check my comment below in the `core_scripts.sh` file

tools/core_scripts.sh
```diff
@@ -0,0 +1 @@
+wget -O tools/ips.py https://raw.githubusercontent.com/runziggurat/ziggurat-core/main/ziggurat-core-scripts/ips.py
```
Rqnsom: some pseudo code for the idea from the `Readme.md` comment:
```
check if ips.py is already downloaded,
   if not download it with wget

call ips.py with all the arguments provided to this script
```

Rqnsom (approved):
LGTM

src/tools/ips.rs
kylegranger: done

src/tools/ips.rs
kylegranger: done

src/tools/ips.rs
kylegranger: done

README.md
kylegranger: For now, I've already moved the one line to the readme.  As for changing the script itself (at this level) I would leave that for Piotr to decide, as this is really new functionality beyond the scope of this PR.  That would be a significant change, including the documentation for the various invocation calls, which would have to be moved into the core_scripts.sh file.  In the future, there may also be other common scripts that we might want to fetch from ziggurat-core.  I prefer orthogonality between the fetching the scripts and the calling of them.

asmie: Let's put it into improvement bucket with _Possible ToDo_ status.
Maybe there will be some other scripts identified which could have similar approach (call & download in one step).

asmie (approved):
LGTM 👍

README.md
kylegranger: I'd say it's a slippery slope, because you could wind up having duplicated scripts like before (albeit wrappers), which was the **original problem we were fixing**.

src/tests/performance/connections.rs
```diff
@@ -141,14 +141,15 @@ async fn p002_connections_load() {
         let node_addr = node.addr();
 
         let mut synth_sockets = Vec::with_capacity(synth_count);
-        let mut ips = IPS.to_vec();
+        // let mut ips = IPS.to_vec();
```
Rqnsom, line 144: let's remove this line in some future PR

README.md
Rqnsom: > original problem we were fixing.

I would say we are discussing two different issues (which can conflict from time to time):

- how to have common scripts in the common external repo and use those as easily as possible
- how to ensure the readme is not clogged with too many instructions _(no user likes massive manuals except for maybe Arch Linux users)_ 😃

the next question is what should

Rqnsom: I'll leave this comment unresolved but will link it from our improvement bucket, we can discuss this more on Slack (it's not a high priority in any way)

src/tests/performance/connections.rs
kylegranger: done

PR #159, opened 2023-04-05
docs: update readme to reflect new destination file for ips.py

kylegranger:
The sample invocations were changed in the ips.py script, but not here.


zeapoz (approved):
LGTM :+1:

PR #160, opened 2023-04-06
chore: cache rust artifacts

zeapoz:


.gitignore
```diff
@@ -1,5 +1,4 @@
 /target
-/Cargo.lock
```
Rqnsom, line 2: can we keep this for normal non-CI users?

Rqnsom (approved):
LGTM

kylegranger (approved):
👍

.gitignore
zeapoz: I guess we could, any particular reason for keeping it?

PR #168, opened 2023-05-25
Call `build-viz-state` workflow.

muzam1l:


.github/workflows/crawler.yml
```diff
@@ -51,8 +51,14 @@ jobs:
           name: latest-result
           path: latest.json
 
-  call-process-results-workflow:
+  call-build-viz-state-workflow:
     needs: [ crawl-network ]
+    uses: runziggurat/ziggurat-core/.github/workflows/build-viz-state.yml@main
+    with:
+      filter: "Ripple"
```
zeapoz, line 58: `"` can be omitted here:
```suggestion
      filter: Ripple
```

zeapoz (approved):
LGTM :+1:

